PR #2, opened 2021-06-25
Selection highlighting between texts and diagrams

christoph-fricke:
This PR adds the options to trigger selection highlighting between the diagram and text. Both possible ways (text selects in diagram, diagram selects in text) can be triggered individually.

I implemented this feature to be part of the `klighd-core` preferences (also appear in the sidebar), even though they are only really useful in VSCode. This makes preferences synchronization with the server for future preferences easier, since this feature already clears the way for the functionality. However, this is up for discussion and might be changed to only be available in VSCode.

## Changes: 
- The `Connection` service has been extended with an `onReady` method, which resolved once the connection is fully setup.
  - Otherwise, preferences (and potentially other information) would be transmitted before they can be processed.
  - Extended existing connection implementations to support this new requirement.
- The preferences registry sends supported preferences (such as the new toggles) to the server using `keith/preferences/setPreferences`.
- New preferences for this feature are added to the general panel and VSCode settings.
- The diagram server now handles `SelectAction` actions.


NiklasRentzCAU (commented):
LGTM. Merge this into main if you want.
I think having the highlighting feature in the core package also makes sense, as for example connecting the standalone tool with an editor would also work out of the box.

christoph-fricke:
:+1:  It would work nearly out of the box. Such tool would have to implement support for the [`diagram.openInTextEditor`](https://github.com/eclipse/sprotty-vscode/blob/master/sprotty-vscode-extension/src/lsp/protocol.ts#L25) LSP notification. It is defined by Sprotty and automatically supported by their Theia and VS Code gluecode. Currently, the standalone tool is unable to handle it in a meaningful way, so it is not implemented. But yeah, everything else required works out of the box.

PR #3, opened 2021-06-26
Master Project - DepthMap & co.

Skgland:
Ports the changes by dwo and our own changes from Keith to klightd-vscode

Still Todo
----
- [x] remove tslint.json
- [x] probably remove packages/klighd-core/.gitignore
- [x] undo change to packages/klighd-core/LICENSE
- [x] check added SetSynthesesAction (probably remove)
- [x] check added SetSynthesesActionData (probably remove)
- [x] check added SetSynthesisAction (probably remove)
- [x] check added SetSynthesisRegistry (probably remove)
- [x] fix RangeOptions currently not working
- [x] client side only RenderOptions
- [x] update UpdateDepthmapModelAction/Command
- [x] expand/collaps is currently not working as expected
- [x] Basic Bookmark functionality
- [x] When RenderOptions with `updateNeeded=false` are changed we need to force a re-render
- [x] Add setting to allow user to disable animating GoToBookmarkActions


packages/klighd-core/src/bookmarks/bookmark-panel.tsx
```diff
@@ -0,0 +1,71 @@
...
+    @inject(DISymbol.RenderOptionsRegistry) private renderOptionsRegistry: RenderOptionsRegistry;
+
+    private static bookmarkActions: [key: string, icon: VNode, action: Action][]
+
+    @postConstruct()
+    init(): void {
+        // Subscribe to different registry changes to make this panel reactive
+        this.synthesesRegistry.onChange(() => this.update());
+        this.preferencesRegistry.onChange(() => this.update());
+        this.renderOptionsRegistry.onChange(() => this.update());
```
christoph-fricke, line 27: Should this panel really synchronize with these registries? Looks like they are not relevant for rendering its UI.

packages/klighd-core/src/bookmarks/bookmark-panel.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+@injectable()
+export class BookmarkPanel extends SidebarPanel {
+    
+    @inject(TYPES.IActionDispatcher) private actionDispatcher: IActionDispatcher;
+    @inject(DISymbol.SynthesesRegistry) private synthesesRegistry: SynthesesRegistry;
+    @inject(DISymbol.PreferencesRegistry) private preferencesRegistry: PreferencesRegistry;
+    @inject(DISymbol.RenderOptionsRegistry) private renderOptionsRegistry: RenderOptionsRegistry;
+
+    private static bookmarkActions: [key: string, icon: VNode, action: Action][]
```
christoph-fricke, line 20: Instead of storing they state in a panel, it should be separated into a `Registry`. Separation between UI and Data. The Registry might be able handle the CreateBookmarkAction directly.

packages/klighd-core/src/bookmarks/bookmark.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+    constructor(@inject(TYPES.Action) protected action: CreateBookmarkAction) {
+        super();
+    }
+
+    execute(context: CommandExecutionContext): CommandReturn {
+        const model = context.root
+        if (isViewport(model)) {
+            BookmarkPanel.addBookmark("MakeBookmark" + BookmarkPanel.getLenghtBookmarks(), <
+                                        div>Bookmark {BookmarkPanel.getLenghtBookmarks()}</div>,
```
christoph-fricke, line 26: Is it necessary to store the UI (JSX) together with the data? This makes it harder to find the right location to change the rendering, which is naturally expected in the panel render method.

packages/klighd-core/src/options/options-module.ts
```diff
@@ -35,6 +36,9 @@ export const optionsModule = new ContainerModule((bind, _, isBound) => {
     bind(GeneralPanel).toSelf().inSingletonScope();
     bind(DISymbol.SidebarPanel).toService(GeneralPanel);
 
+    bind(BookmarkPanel).toSelf().inSingletonScope();
+    bind(DISymbol.SidebarPanel).toService(BookmarkPanel);
+
```
christoph-fricke, line 41: The panel should be registered in the bookmark module, since it belongs to the module.

christoph-fricke (changes requested):

packages/klighd-core/src/bookmarks/bookmark-panel.tsx
Skgland: added a BookmarkRegistry and only listen for changes there now

packages/klighd-core/src/bookmarks/bookmark-panel.tsx
Skgland: added a BookmarkRegistry, but I don't think the Registry can handle the Action directly as it need access to the model which I don't know how to access from the Action in thee registry

packages/klighd-core/src/bookmarks/bookmark.tsx
Skgland: I think this is also now solved by thee BookmarkRegistry

packages/klighd-core/src/options/options-module.ts
Skgland: has been moved to the bookmark module

packages/klighd-core/src/bookmarks/bookmark-panel.tsx
christoph-fricke: @Skgland My thought was that the command might dispatch an extra add action with the model attached. However, I am not sure whether or not a command is supposed to trigger further actions. The Sprotty architecture overview suggests otherwise (?). The way you handled it in the new commit is properly the best call.

packages/klighd-core/src/bookmarks/bookmark-panel.tsx
```diff
@@ -0,0 +1,51 @@
...
+        this.bookmarkRegistry.bookmarks.forEach((bookmark,index) => {
+            list.push(
+                <div>
+                    <button on-click={() => this.handleBookmarkActionClick(bookmark)}>{bookmark.name ?? ("Bookmark " + index)}</button>
+                </div>
+            )
+        })
+
+        return  <div>{...list}</div>;
+    }
```
christoph-fricke, line 45:
```suggestion
    render(): VNode {
        return <div>{this.bookmarkRegistry.bookmarks.map((bookmark,index) => (
                <div key={index}>
                    <button on-click={() => this.handleBookmarkActionClick(bookmark)}>{bookmark.name ?? ("Bookmark " + index)}</button>
                </div>
            )
        })}</div>;
    } 
```

I think a function approach fits better in this scenario. No need to store the VNodes explicitly in an extra array.

packages/klighd-core/src/depth-map.ts
```diff
@@ -0,0 +1,483 @@
...
+
+        // Check all collected child regions of expanded states.
+        childSet.forEach(childRegion => {
+            const vis = this.getExpansionState(childRegion, viewport, threshold);
+            if (vis === Visibility.Expanded) {
+                this.searchUntilCollapse(childRegion, viewport, threshold)
+            } else {
+                if (childRegion.parent) {
+                    // our parent is expanded and we are not, 
+                    // this meand our parent is a critical region
```
NiklasRentzCAU, line 354: ~~meand~~ means

packages/klighd-core/src/depth-map.ts
```diff
@@ -0,0 +1,483 @@
...
+     * @returns The appropriate expansion state.
+     */
+    getExpansionState(region: Region, viewport: Viewport, threshold: number): Visibility {
+        // Collapse all invisible states and regions.
+        if (!this.isVisible(region, viewport)) {
+            return Visibility.OutOfBounds
+        } else if (!region.parent) {
+            // Regions without parents should always be expanded if they are visible
+            return Visibility.Expanded
+        } {
```
NiklasRentzCAU, line 380: Is there an else missing? Or is this the intended behavior? Either way, just opening a scope like this feels like a bug.

packages/klighd-core/src/depth-map.ts
```diff
@@ -0,0 +1,483 @@
...
+            return true
+        }
+    }
+
+    /** 
+     * Compares the size of a node to the viewport and returns the smallest fraction of either height or width.
+     * 
+     * @param node The KNode in question
+     * @param viewport The curent viewport
+     * @returns
```
NiklasRentzCAU, line 418: empty tag

packages/klighd-core/src/options/components/icons.tsx
```diff
@@ -98,3 +98,19 @@ export function ExportIcon(): VNode {
         </svg>
     );
 }
+
+export function  BookmarkIcon(): VNode {
```
NiklasRentzCAU, line 102: double whitespace

packages/klighd-core/src/options/render-options-registry.ts
```diff
@@ -16,18 +16,147 @@ import { injectable } from "inversify";
...
+ * Threshold for expansion as well as collapse of states.
+ * Corresponds to the regions size compared to the current viewport.
+ */
+ export class ExpandCollapseThreshold implements RangeOption {
+    static readonly ID: string = 'expand-collapse-threshold'
+    static readonly NAME: string = 'Expand Collapse Threshold'
+    readonly id: string = ExpandCollapseThreshold.ID
+    readonly name: string = ExpandCollapseThreshold.NAME
+    readonly type: TransformationOptionType = TransformationOptionType.RANGE
+    readonly updateNeeded: boolean = false
```
NiklasRentzCAU, line 58: Why is this field added if it is always false? Especially this option seems to benefit if changing the threshold would also live update the diagram accordingly.

packages/klighd-core/src/update/update-depthmap-model.ts
```diff
@@ -0,0 +1,52 @@
...
+        if ((input as SModelRootSchema).id !== undefined)
+            this.newRoot = input as SModelRootSchema;
+    }
+}
+
+/**
+ * UpdateModelCommand gets fired whenever a setModel or updateModel was executed
+ */
+@injectable()
+export class UpdateDepthmapModelCommand extends Command{
```
NiklasRentzCAU, line 32: missing whitespace before `{`

packages/klighd-core/src/views-rendering.tsx
```diff
@@ -106,10 +95,10 @@ export function renderRectangularShape(rendering: KContainerRendering, parent: S
         colorStyles.background = DEFAULT_CLICKABLE_FILL
     }
     const shadowStyles = getSvgShadowStyles(styles, context)
-    const lineStyles = getSvgLineStyles(styles)
+    const lineStyles = getSvgLineStyles(styles,parent,context)
```
NiklasRentzCAU, line 98: whitespace after each ,

packages/klighd-core/src/views-rendering.tsx
```diff
@@ -266,6 +255,41 @@ export function renderRectangularShape(rendering: KContainerRendering, parent: S
...
+        const region = context.depthMap.getRegion((parent as KNode).id)
+        if (region && region.expansionState !== Visibility.Expanded && !region.hasTitle) {
+            // Render indirect region titles.
+            if (region.superStateTitle) {
+                const titleSVG = renderKText(region.superStateTitle, region.boundingRectangle, propagatedStyles, context, mListener)
+                element.children ? element.children.push(titleSVG) : element.children = [titleSVG]
+            } else if (region.macroStateTitle && !region.hasMultipleMacroStates) {
+                const titleSVG = renderKText(region.macroStateTitle, region.boundingRectangle, propagatedStyles, context, mListener)
+                element.children ? element.children.push(titleSVG) : element.children = [titleSVG]
+                // If there is no title draw placeholder.
```
NiklasRentzCAU, line 268: I think the placeholder could be rendered always, if there is further content to be expanded, otherwise this seems insonsistent if only some regions have that symbol.

packages/klighd-core/src/views-rendering.tsx
```diff
@@ -518,6 +568,43 @@ export function renderKText(rendering: KText, parent: SKGraphElement | SKLabel,
             attrs.textLength = rendering.calculatedTextLineWidths[0]
             attrs.lengthAdjust = 'spacingAndGlyphs'
         }
+
+        if (context.depthMap) {
+            if (boundsAndTransformation.bounds.width && boundsAndTransformation.bounds.height
+                && (rendering.isNodeTitle || (text !== '-' && text !== '+'))) {
```
NiklasRentzCAU, line 574: This seems to be a hack to avoid some behavior that happens with SCCharts, this code however should be generic for any synthesis. Is there a way to avoid this?

packages/klighd-core/src/views-styles.tsx
```diff
@@ -345,6 +346,7 @@ export function applyKStyle(style: KStyle, styles: KStyles, stylesToPropagage?:
  * @param from The KStyles to copy from.
  * @param to The KStyles to copy to.
  */
+
```
NiklasRentzCAU, line 349: superfluous empty line

packages/klighd-core/src/views-styles.tsx
```diff
@@ -624,8 +626,22 @@ export function isInvisible(styles: KStyles): boolean {
  * 'stroke-miterlimit' has to be set to the miterLimit style. (This is not a string, but a number.)
  * @param styles The KStyles of the rendering.
  */
-export function getSvgLineStyles(styles: KStyles): LineStyles {
-    const lineWidth = styles.kLineWidth === undefined ? DEFAULT_LINE_WIDTH : styles.kLineWidth.lineWidth
+export function getSvgLineStyles(styles: KStyles, target:SKGraphElement, context: SKGraphModelRenderer): LineStyles {
```
NiklasRentzCAU, line 629: tags for new params missing.

packages/klighd-core/src/views.tsx
```diff
@@ -107,24 +168,14 @@ export class KNodeView implements IView {
...
-            if (interactiveNodes) {
-                result.push(interactiveNodes)
-            }
-            result.push(...children)
+        let shadow = undefined
 
-            return <g>{...result}</g>
+        if (isShadow) {
+            // Render shadow of the node
+            shadow = getRendering(node.data, node, new KStyles, ctx, this.mListener)
```
NiklasRentzCAU, line 178: what are these changes for? Looks like some lines switched positions here.

NiklasRentzCAU (changes requested):
Most changes seem fine and fit into the code well. Just care about the comments I left in the code changes.

packages/klighd-core/src/depth-map.ts
```diff
@@ -0,0 +1,483 @@
...
+        let current = node.parent as KNode;
+        while (current) {
+
+            const region = this.getRegion(current.id);
+
+            if (region) {
+                return region
+            }
+
+            current = current.parent as KNode
```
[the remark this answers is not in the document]
Skgland: Yes, it is possible for there to not be a Region that contains a node, though this should only be the case for the root nodes.

packages/klighd-core/src/views.tsx
Skgland: yes, that is basically what is done here.
this is done so that we don't need to get the rendering for the root node which would just be discarded in that case

packages/klighd-core/src/options/render-options-registry.ts
Skgland: * This is `true` for the `ShowConstraintOption` above in Line 27.
*  Client Side the Diagram should still be updated (TODO), but this does not require the LS to regenerate the model

packages/klighd-core/src/bookmarks/bookmark.ts
```diff
@@ -0,0 +1,89 @@
...
+    readonly kind = GoToBookmarkAction.KIND;
+    
+    viewport: Viewport;
+    
+    constructor(bookmark: Bookmark){
+        this.viewport = bookmark.place;
+    }
+}
+
+export class GoToBookmarkCommand extends Command {
```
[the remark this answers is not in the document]
Skgland: the handler for the GoToBookmark Action now delegated to the SetViewportAction instead which made the GoToBookmarkCommand obsolete

packages/klighd-core/src/depth-map.ts
Skgland: Yes an else is missing though as all previous branches diverge by returning this has the same behaviour

packages/klighd-core/src/bookmarks/bookmark-module.ts
```diff
@@ -0,0 +1,32 @@
+import { ContainerModule } from "inversify";
```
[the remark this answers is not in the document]
Skgland: I think I have now updated all our added files in this regard.

The header references Kieler, but this coded is now under klighd-vscode, should the header be adjusted accordingly?

packages/klighd-core/src/views-rendering.tsx
Skgland: I don't know how to correctly identify the text elements we want to handle here and as is it appears to have plenty of false positives.

What we would want is all Node Titles and all Region Titles, the former appears to be covered correctly, but I don't know how to identify the latter.

Skgland: My latest commits here in combination with the enabling commits to klighd and semantics should resolve this

packages/klighd-core/src/views-rendering.tsx
Skgland: I believe this should be resolved by Mikas latest changes

packages/klighd-core/src/depth-map.ts
```diff
@@ -0,0 +1,483 @@
...
+                    height: (child as KNode).bounds.height
+                }
+                nextRegion.parent = region
+                region.children.push(nextRegion)
+                // In the models parent child structure a child can occur multiple times.
+                this.addRegionToMap(nextRegion)
+                // Continue with the children of the new region.
+                this.initHelper((child as KNode), (depth + 1), nextRegion, nextRegion.absoluteBounds.x, nextRegion.absoluteBounds.y)
+            } else {
+                // Continue with the other children on the current depth level.
```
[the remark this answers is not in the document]
Skgland: Each nodes provided/containing Region is now initialized when the first of them is accessed.

packages/klighd-core/src/views.tsx
Skgland: this change is now reverted as apparently the root node depends on a side-effect of the getRendering call when rendering elkt diagrams

packages/klighd-core/src/options/render-options-registry.ts
Skgland: this field has now been removed again as you are right that it is not necessary as it can be false for all RenderOptions

packages/klighd-core/src/views-rendering.tsx
Skgland: as the former turned out to be not necessary and the later has been merged this should by now be resolved

packages/klighd-core/src/bookmarks/bookmark-panel.tsx
```diff
@@ -0,0 +1,228 @@
...
+                    >
+                        <i attrs={{ "data-feather": "check" }} />
+                    </button>
+                    <input type="text" value={bookmark.name ?? ""} placeholder="Enter Bookmark Name" on-keypress={(event: KeyboardEvent) => this.inputSaveBookmarkName(bookmark, event)} />
+                </span>
+            </fieldset>
+        </div>
+    }
+
+    private handleBookmarkGoto(bookmark: Bookmark) {
```
NiklasRentzCAU, line 131: You configured the registry to be configurable by injection, for that to work properly the methods should all be protected here, so that someone extending or altering the behavior can do so in an easy fashion.

packages/klighd-core/src/views-rendering.tsx
```diff
@@ -477,6 +489,37 @@ export function renderKText(rendering: KText, parent: SKGraphElement | SKLabel,
...
     const textStyles = getSvgTextStyles(styles)
 
+    // Replace text with rectangle, if the text is too small.
+    const region = context.depthMap?.getProvidingRegion(parent as KNode, context.viewport, context.renderingOptions)
+
+    const simplifySmallTextOption = context.renderingOptions.getValueForId(SimplifySmallText.ID)
+    const simplifySmallText = simplifySmallTextOption ?? false // Only enable, if option is found.
+    if (simplifySmallText && (!region || region.detail === DetailLevel.FullDetails) && !rendering.isNodeTitle) {
+        const simplificationThresholdOption = context.renderingOptions.getValueForId(TextSimplificationThreshold.ID)
+        const defaultThreshold = 3
```
NiklasRentzCAU, line 499: defaults should be stored in a more central location and not as some magic number in the rendering code. For this example maybe as TextSimplificationThreshold.DEFAULT

packages/klighd-core/src/views-rendering.tsx
```diff
@@ -522,11 +565,80 @@ export function renderKText(rendering: KText, parent: SKGraphElement | SKLabel,
             attrs.textLength = rendering.calculatedTextLineWidths[0]
             attrs.lengthAdjust = 'spacingAndGlyphs'
         }
-        elements = [
-            <text {...attrs}>
-                {...children}
-            </text>
-        ]
+        const defaultOverlayThreshold = 4
```
NiklasRentzCAU, line 568: See above.

NiklasRentzCAU (changes requested):

christoph-fricke (changes requested):
Really cool to see the my ideas already applied to other work as well! Looks like they are some useful abstractions. :smile: 

I noticed a few things that might be improved. The one suggestion that absolutely must be applied is the relative import across package boundaries.

applications/klighd-cli/src/main.ts
```diff
@@ -24,11 +24,13 @@ import {
...
     SetPreferencesAction,
     bindServices,
+    SetInitialBookmark,
 } from "@kieler/klighd-core";
 import { LSPConnection } from "./services/connection";
-import { getDiagramSourceUri, getLanguageId, readSearchParam, sleep } from "./helpers";
+import { getBookmarkViewport, getDiagramSourceUri, getLanguageId, readSearchParam, sleep } from "./helpers";
 import { showSpinner, hideSpinner } from "./spinner";
 import { showPopup } from "./popup";
+import { Bookmark } from "../../../packages/klighd-core/lib/bookmarks/bookmark";
```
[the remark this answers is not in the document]
Skgland: I believe I have now fixed all of the incorrect cross package imports in commit c2e04a7a

packages/klighd-core/src/bookmarks/bookmark-registry.ts
```diff
@@ -0,0 +1,82 @@
...
+        if (bm) {
+            if (new_name === "") {
+                bm.name = undefined
+            } else {
+                bm.name = new_name
+            }
+            this.notifyListeners();
+        }
+
+    }
```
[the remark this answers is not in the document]
Skgland: Adjusted in commit 56468f69 to have all changes now go through the dispatcher, not sure if the CreateBookmarkCommand is ideal this way. I couldn't find a reasonable way to dispatch an AddBookmarkAction from the Command

applications/klighd-cli/src/helpers.ts
```diff
@@ -45,3 +47,22 @@ export function getLanguageId(documentUri: string): string {
...
+        const zoom = parseFloat(zoomParam);
+
+        if (!(isNaN(x) || isNaN(y) || isNaN(zoom))) {
+            return { scroll: { x: x, y: y }, zoom: zoom }
+        }
+    }
+
+    return null
+
+}
```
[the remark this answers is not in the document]
Skgland: The whole function can't be moved to `klighd-core` as `readSearchParameter` is only available in the CLI.
Should I move just the object construction and NaN check to core or also the floatParsing and the null check?
And where in core would be appropriate?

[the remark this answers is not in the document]
christoph-fricke: Your are properly right. In the core, it should not have all this parameter checking and "just" expect a number. So nothing would really change, since it mostly consists of massaging parameter...

I would have put it into https://github.com/kieler/klighd-vscode/blob/main/packages/klighd-core/src/base/external-helpers.ts.

packages/klighd-core/src/bookmarks/bookmark-registry.ts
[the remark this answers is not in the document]
christoph-fricke: I see. The command can inject the `TYPES.ActionDispatcher` to dispatch the action. Commands should properly be pure operations and not have side-effects :thinking: . However, I know why the commands exist and I don't know a better way as well... Is it possible to dispatch Create/Delete actions in the undo/redo methods of the command as well? It would make the command more complete and overall bullet proof.

Also, a little pro-tip: You can add [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) to the actions to avoid the necessary casting when they are handled. Instead of comparing the action kind in the "if statement", it would be compared in the type predicate.
The [actions for options](https://github.com/kieler/klighd-vscode/blob/main/packages/klighd-core/src/options/actions.ts) should be a good example for such type predicate.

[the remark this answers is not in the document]
Skgland: Changed CreateBookmarkCommand to have the ActionDispatcher injected and use that to dispactch an AddBookmarkAction, also implemented undo and redo. 

Type predicates for the Actions have been added and are now used

applications/klighd-cli/src/helpers.ts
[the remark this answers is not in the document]
Skgland: Ok, moved the check and construction to `external-helper.ts` in core,.
Btw. changed the !isNaN checks to isFinite check as Infinity values should also not be accepted.

christoph-fricke (approved):

NiklasRentzCAU (approved):
In its current state this looks good to me. @christoph-fricke I would say we wait with merging this though so you and the project can have their end presentations with their own code base. We will merge this after you all are done with the semester.
Note: merging this will also require switching from a Language Server based on the latest KLighD release to a LS based on the current KLighD master due to a change in child area placement.

Skgland:
Reopening as #19 did not merge to main

PR #7, opened 2021-07-14
Add Open VSX as distribution target

christoph-fricke:
Extends the distribution script for the `klighd-vscode` package to also publish to Open VSX.

To save a few CI seconds, the script is further changed to not package the extension. Instead, the extension is only packaged once before it is distributed. Previously, a call to `vsce publish` would have also packaged the extension.


NiklasRentzCAU (approved):
LGTM.

Check https://github.com/EclipseFdn/open-vsx.org/issues/585 to see whenever publishing in the Open VSX Registry under the kieler namespace is possible.

PR #11, opened 2021-08-12
Core: Fix layout used by text input options (#10)

### LINKED ISSUE #10: Hideous layout for text options
Text options rendered in the options sidebar look rather hideous. Their input field and label are next to each other. Instead, the label text should be above the input. Similar to range options. Furthermore, the input fields could use a bit of theming. They do not work well with a dark theme in VS Code.

Example screenshot is provided by the [OSGiViz VS Code extension](https://marketplace.visualstudio.com/items?itemName=kieler.osgiviz):

image

christoph-fricke:
This fixes the issues identified in #10. Text inputs now use a vertical layout with the label placed on top. Furthermore, they are styled using the available theme, similar to the synthesis selector.

The issue (#10) contains a screenshot from before the fix. After the fix:
image

closes #10


NiklasRentzCAU (approved):
LGTM.

PR #12, opened 2021-08-17
Feat: Add the ability to persist data

christoph-fricke:
This PR adds the long awaited ability to persist data between reloads. The data is send to the LS during initialization under the `clientDiagramOptions` key (same as before).

To achieve this, a `PersistenceStorage` service is added to the core. An `OptionsPersistence` action handler handles all actions that update diagram options (render, layout, synthesis) and persists them using the new service.

Each platform provides their implementation for the service and is responsible for notifying the server about the data during initialization. 

The CLI uses the `localStorage` to persist the data long term. A URL search param and a CLI option have been added for the user to nuke persisted data. Since `localStorage` depends on the origin, which is port dependent, the CLI has been switched from a random port to a series of preferred ports (falls back to a random port after 6 availability tests).

The VS Code extension uses messages to sync the data between the webview and extension. The data is stored in the workspace state, which means that it will change between opened folder. Say a user has two projects, they would use different persisted options. @NiklasRentzCAU  This is up to debate and can be change easily to a global storage, because both use the [`Memento`](https://code.visualstudio.com/api/references/vscode-api#Memento) API.
An command has been added to the extension as well, which allows the user to nuke persisted data. I think it is quite useful if one wants to return to the defaults, but it only truly applies to the LS after a reload (since the LS is only initialized on startup...).


applications/klighd-vscode/src/extension.ts
```diff
@@ -63,6 +70,18 @@ export function activate(context: vscode.ExtensionContext): void {
...
     );
 
+    // Command for the user to remove all data stored by this extension. Allows
+    // the user to reset changed synthesis options etc.
+    context.subscriptions.push(
+        vscode.commands.registerCommand(command.clearData, () => {
+            StorageService.clearAll(mementoForPersistence);
+            // Inform the user that the deletion does not effect previously initialized options on the LS.
+            vscode.window.showInformationMessage(
+                "Stored data has been deleted. Diagram options will be reset to default after a reload."
```
NiklasRentzCAU, line 80: This could also notify the core that all options should be reset to their default value right now, making a window reload obsolete.

packages/klighd-core/README.md
```diff
@@ -51,16 +61,19 @@ import {
     SetPreferencesAction,
     bindServices,
 } from "@kieler/klighd-core";
-// Your implementation of the `Connection` interface.
+// Your implementation of the `Connection` interface
 import { ConnectionImpl } from "./services/connection";
+// Your implementation of the `PersistenceStorage` interface
```
NiklasRentzCAU, line 66: I think this should link to the default implementation, if nothing should be changed for the specific extension.

NiklasRentzCAU (changes requested):
Looks good, but I left two comments to answer before merging

packages/klighd-core/README.md
christoph-fricke: What default implementation? There exists no default implementation in `klighd-core`. This is supposed to provide an idea about how an integration of `klighd-core` can be approached.

NiklasRentzCAU: I thought this was the readme for the extension, where a default implementation should be available. Did not read the file path correctly during review.

applications/klighd-vscode/src/extension.ts
christoph-fricke: This *tiny* improvement only required changes to 13 files. Hopefully users notice it :D

*insert "this is fine" meme*

applications/klighd-vscode/src/extension.ts
```diff
@@ -75,10 +76,17 @@ export function activate(context: vscode.ExtensionContext): void {
     context.subscriptions.push(
         vscode.commands.registerCommand(command.clearData, () => {
             StorageService.clearAll(mementoForPersistence);
-            // Inform the user that the deletion does not effect previously initialized options on the LS.
-            vscode.window.showInformationMessage(
-                "Stored data has been deleted. Diagram options will be reset to default after a reload."
-            );
+            // webviews are manages by the SprottyExtensions which in return are stored by this extension in a lookup map
```
NiklasRentzCAU, line 79: *managed

applications/klighd-vscode/src/storage/storage-service.ts
```diff
@@ -65,14 +65,31 @@ export class StorageService {
...
      * The data in this service should be the source of truth.
      */
     private reportData(send: Send, data: Record<string, any>) {
         send({ type: "persistence/reportItems", payload: { items: data } });
     }
 
+    /**
+     * Reports a clear back to the webview.
+     *
+     * _NOTE_: At the moment, this will never be reported as it can only reports for
```
NiklasRentzCAU, line 79: *report

NiklasRentzCAU (approved):
*This **is** fine.*

(Except for those two typos)

PR #13, opened 2021-08-19
Feat: Add sync with editor to the VS Code extension

christoph-fricke:
The VS Code extension is now able to disable "sync with editor", meaning the diagram view no longer updates when the active editor changes and it is disabled.

The behavior should be identical to the implementation in KEITH, with the exception of the option location. In KEITH it was part of the editor menu. Now it is part of the editor toolbar, because I was unable to figure out how to add a check icon next to a menu entry... Also the icon changed from a link/check icon to a "sync" icon,  because it has a nice "sync-ignore" counterpart, which provides nice visual feedback to the user.

Lastly, in KEITH a user is not able to change the model for the diagram view when "sync with editor" is disabled. I think the ability to do this is important and nice to have, so I found a way to allow a user to open a diagram while  "sync with editor" is disabled.


applications/klighd-vscode/src/klighd-webview.ts
```diff
@@ -49,4 +81,18 @@ export class KLighDWebview extends SprottyLspWebview {
             })
         );
     }
+
+    /** Toggles "sync with editor". If disabled, the diagram view will not update when the active editor changes. */
+    toggleSyncWithEditor(sync: boolean): void {
```
NiklasRentzCAU, line 86: This method name is a bit misleading: a toggle function should not have a parameter as just toggling flips the option. I guess this should rather be named `setSyncWithEditor`.

NiklasRentzCAU (changes requested):
Other than that one minor detail, LGTM.

applications/klighd-vscode/src/klighd-webview.ts
christoph-fricke: Refactored.

PR #14, opened 2021-09-02
Feat: Add `serve` command to the CLI

christoph-fricke:
This PR adds a new `serve` command and improves the overall UX of the CLI. All improvements/changes are listed below.
The `serve` command starts the web server without requiring a file path and opening the view. This command is especially useful for diagram views embedded in local documentation, since it has no other side-effects apart from starting the web server, and can be used in scripts.

- Added `serve` command.
- Moved previous behavior to an `open` command.
  - => Default command that will be used if no command is provided.
- The CLI version is no parsed from the `package.json` version when it is packaged. No need to update it manually...
- Added a global `--port` / `-p` option to overwrite the port used for the web server.
- Commands and options are now sorted in the help output.
- Altered descriptions for commands and options to be shorter and better align with the text of the auto generated help.
- The script name (used in the help) is now parsed dynamically  (`process.title`) to reflect the name of the executable.
  - When a relative path is used to call the executable, the relative path is reflected in the script name.
  - Should avoid confusion in the help text if the executable is named differently. Especially since they are packaged with their OS target appended.
- Updated Commander to the latest major version


NiklasRentzCAU (approved):
LGTM!

PR #16, opened 2021-09-15
Docs: Add a local documentation example

christoph-fricke:
This PR adds an example and usage guide on "Using the CLI for Local Documentation".

Sadly, before a user is able to try the example, the diagram-file paths in the html have to be modified to match the local file structure. I tried to use relative paths hoping that it would work if the CLI is used in the same folder, but it generated an exception on the LS.

### DIFF
--- a/examples/local-documentation/README.md
+++ b/examples/local-documentation/README.md
@@ -1,0 +1,24 @@
+# Using the CLI for Local Documentation
+
+This example highlights how the CLI can be used to embed a diagram visualization in local web-based
+documentation.
+
+> The key ingredients are `iframe` elements and the `klighd serve` command.
+
+`Serve` starts the web server, but does not open a diagram view. Instead, `iframe` elements embed
+diagram views into the documentation by pointing to the started web server and provide paths to the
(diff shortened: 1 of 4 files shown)


NiklasRentzCAU (approved):
Nice example. I tested this to see how this would for me, and only one of the two embedded diagrams loaded for me. Did you change anything on the server side to work with both diagrams as well?

christoph-fricke:
> Nice example. I tested this to see how this would for me, and only one of the two embedded diagrams loaded for me. Did you change anything on the server side to work with both diagrams as well?

That is strange. Works fine for me. Did you update the `source` paths for both iframe elements? I do not see a reason why it would not work otherwise.

christoph-fricke:
Note for anyone that stumbles upon this PR:

The example does not work when the language server (LS) is started in development mode and connected to on a port. The single server is not able to handle multiple connections. When a JAR is used, the web server is able to spawn multiple LS for multiple connections, as it is the case in this example.

Therefore, **to try the example a LS JAR has to be used**.

PR #29, opened 2021-12-06
Store preferences

soerendomroes:
Migrated resize to fit, use light background, and animate bookmark to render-options-registry.
Removed initial vscode extension settings.

Added clear handler for preferences.

See https://rtsys.informatik.uni-kiel.de/jira/browse/KVC-13


applications/klighd-cli/src/main.ts
```diff
@@ -22,11 +22,11 @@ import {
     createKlighdDiagramContainer,
     requestModel,
     getActionDispatcher,
-    SetPreferencesAction,
     bindServices,
     SetInitialBookmark,
     Bookmark,
 } from "@kieler/klighd-core";
+import { SetPreferencesAction } from "@kieler/klighd-core/lib/options/actions"
```
NiklasRentzCAU, line 29: For the API this should be importable as before from `@kieler/klighd-core` without the `lib/...` part after it.

NiklasRentzCAU (changes requested):
Looks good to unify the preferences into the options.
However, the 'use light background' option does not work anymore in the klighd-cli with this PR.

NiklasRentzCAU:
The light background works as intended, I was just confused for some reason. Please delete the branch if not needed anymore.

Issue #28, opened 2021-12-02
Node titles with multiple stiched together KText elements create unwanted rendering of the titles when zoomed out


NiklasRentzCAU:
If some synthesis chooses to mark multiple KText elements per node as "node title", such as SCCharts does for class hierarchies and referenced charts, the rendering goes wrong: 
![overlapping-node-label](https://user-images.githubusercontent.com/47597619/144458695-47237233-6684-4391-b3be-d4e4e7408f2b.png)

Initially the feature expected at most one KText per node to have this property and this case shows that this cannot be expected.

NiklasRentzCAU:
I think this may be solvable by changing the node title KText to a single node title general rendering which will be enlarged as a single text is currently. That way for this case the (invisible) rectangle containing these texts can be marked as the node 'label' and will therefore be enlarged to fit the space.

PR #30, opened 2021-12-09
Node titles allow for any rendering, not only KTexts

NiklasRentzCAU:
Initially node titles were implemented specific to KTexts, so that any KText could be a node title which would be scaled up within its parent rendering when zoomed out far enough to make that title illegible.
This PR rethinks the core idea and allows the same concept for any rendering. So now any rendering can be set to be the node title and that entire rendering (e.g. formatted text) will be scaled accordingly.

Also re-organizes some code used multiple times.


NiklasRentzCAU:
Closes #28.

PR #31, opened 2021-12-18
Small bug-fix to successfully run 'yarn package'

CheckerFlow:
I tried to generate the packages by running 'yarn package' but got an error. 
Fixed the error by removing some presumably unnecessary quotes.


NiklasRentzCAU:
On my system (Linux) the command as it is currently will write `exports.VERSION = "0.2.1";` to the versions.js file whereas your change will cause it to error and write `exports.VERSION = ""$npm_package_version"";` to the same file. What OS and yarn version are you using?

The GH actions build being green is not to be trusted here, as it does not package the project, it only builds it.

PR #60, opened 2022-02-21
Reuse SModel index for diagram pieces

NiklasRentzCAU:
recalculate LOD when new diagram pieces arrive. Moved diagram piece code together in a single folder

Moves the diagram-piece-related actions into its folder, removes now-invalid assumption that a no-pan or -zoom update does not need to update the depth map and replaces a single line in the diagram piece request to not create the model anew on the client for each update:
`this.root = context.modelFactory.createRoot(context.root)` to `this.root = context.root`


Eddykasp (approved):
I see no problems

PR #63, opened 2022-03-22
update properties to be accessed using <string,unknown> map

Eddykasp:
This changes the way elk properties are accessed by the client.
Instead of the NodeProperties class, properties are now a map<string, unknown>, where the key is the the fully qualified name such as "org.eclipse.elk.algorithm"
Because the property types are now unknown, they have to be explicitly cast by the developer meaning the developer has to know the type of the property. Currently all the used properties are primitive types, but more complex types also exist and may need to be dealt with nicely. There may be some more checks necessary to ensure non-existent properties aren't being accessed, but I haven't found any such cases yet. 

[KGraphMappingUtil](https://github.com/kieler/KLighD/blob/72d79588917b5ef7c0e7b970c380d1584dd97419/plugins/de.cau.cs.kieler.klighd.lsp/src/de/cau/cs/kieler/klighd/lsp/utils/KGraphMappingUtil.xtend#L134) shows how the properties are being set in KlighD. This needs to be improved a bit and the duplicate properties being set in KGraphDiagramGenerator still need to be removed.

### DIFF
--- a/packages/klighd-core/src/views-rendering.tsx
+++ b/packages/klighd-core/src/views-rendering.tsx
@@ -1065,8 +1065,8 @@ export function getKRendering(datas: KGraphData[], context: SKGraphModelRenderer
                 const id = (data as KRenderingRef).properties['klighd.lsp.rendering.id'] as string
                 for (const rendering of context.kRenderingLibrary.renderings) {
                     if ((rendering as KRendering).properties['klighd.lsp.rendering.id'] as string === id) {
-                        context.boundsMap = (data as KRenderingRef).calculatedBoundsMap
-                        context.decorationMap = (data as KRenderingRef).calculatedDecorationMap
+                        context.boundsMap = (data as KRenderingRef).properties['klighd.lsp.calculated.bounds.map']
+                        context.decorationMap = (data as KRenderingRef).properties['klighd.lsp.calculated.decoration.map']
                         return rendering as KRendering
                     }
                 }
--- a/packages/klighd-core/src/views-rendering.tsx
+++ b/packages/klighd-core/src/views-rendering.tsx
@@ -50,7 +50,7 @@ export function renderChildArea(rendering: KChildArea, parent: SKGraphElement, p
     // remember, that this parent's children are now already rendered
     parent.areChildAreaChildrenRendered = true
 
-    const element = <g id={rendering.renderingId}>
+    const element = <g id={rendering.properties['klighd.lsp.rendering.id'] as string}>
         {context.renderChildAreaChildren(parent)}
     </g>
 
@@ -153,7 +153,7 @@ export function renderRectangularShape(
                     }
                 }
 
-                element = <g id={rendering.renderingId} {...gAttrs}>
+                element = <g id={rendering.properties['klighd.lsp.rendering.id'] as string} {...gAttrs}>
                     {...renderSVGArc(lineStyles, colorStyles, shadowStyles, d, styles.kShadow)}
                     {renderChildRenderings(rendering, parent, stylesToPropagate, context, childOfNodeTitle)}
                 </g>
@@ -164,7 +164,7 @@ export function renderRectangularShape(
         }
         // eslint-disable-next-line
         case K_ELLIPSE: {
-            element = <g id={rendering.renderingId} {...gAttrs}>
+            element = <g id={rendering.properties['klighd.lsp.rendering.id'] as string} {...gAttrs}>
                 {...renderSVGEllipse(boundsAndTransformation.bounds, lineStyles, colorStyles, shadowStyles, styles.kShadow)}
                 {renderChildRenderings(rendering, parent, stylesToPropagate, context, childOfNodeTitle)}
             </g>
@@ -179,7 +179,7 @@ export function renderRectangularShape(
             const rx = (rendering as KRoundedRectangle).cornerWidth
             const ry = (rendering as KRoundedRectangle).cornerHeight
 
-            element = <g id={rendering.renderingId} {...gAttrs}>
+            element = <g id={rendering.properties['klighd.lsp.rendering.id'] as string} {...gAttrs}>
                 {...renderSVGRect(boundsAndTransformation.bounds, rx, ry, lineStyles, colorStyles, shadowStyles, styles.kShadow)}
                 {renderChildRenderings(rendering, parent, stylesToPropagate, context, childOfNodeTitle)}
             </g>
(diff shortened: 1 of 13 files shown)


packages/klighd-interactive/src/constraint-classes.ts
```diff
@@ -52,10 +52,11 @@ export class KNode extends RectangularNode implements KGraphElement {
     areChildAreaChildrenRendered = false
     areNonChildAreaChildrenRendered = false
     hasFeature(feature: symbol): boolean {
-        return feature === selectFeature || (feature === moveFeature && (this.parent as KNode).properties.interactiveLayout)
+        return feature === selectFeature || (feature === moveFeature && (this.parent as KNode).properties.get("org.eclipse.elk.interactiveLayout") as boolean)
     }
 
-    properties: NodeProperties
+    //properties: NodeProperties
+    properties: Map<string, unknown>
```
NiklasRentzCAU, line 59: You could throw a bit of Typescript's almighty type system on this map to make using it with known keys to not need the `as number` or similar casts every time a property is used outside of equality checks. See for example [this commit](https://github.com/kieler/vscode/commit/25ca28bf430ba31ec0bedb7c2daf49cde55ebd0f) on kieler/vscode.

packages/klighd-interactive/src/constraint-classes.ts
```diff
@@ -52,10 +52,11 @@ export class KNode extends RectangularNode implements KGraphElement {
     areChildAreaChildrenRendered = false
     areNonChildAreaChildrenRendered = false
     hasFeature(feature: symbol): boolean {
-        return feature === selectFeature || (feature === moveFeature && (this.parent as KNode).properties.interactiveLayout)
+        return feature === selectFeature || (feature === moveFeature && (this.parent as KNode).properties.get("org.eclipse.elk.interactiveLayout") as boolean)
```
NiklasRentzCAU, line 55: We try to use single quotes for strings in this project. This is, however, not enforced (yet).

packages/klighd-interactive/src/constraint-classes.ts
NiklasRentzCAU: Having the keys of all properties as 'magic strings' may also not be ideal here and may cause bugs due to spelling errors. Having these as static variables to refer to somewhere or allow the type system to only allow correct keys as within the above commit could help here.

NiklasRentzCAU (commented):
All the now necessary casts and exact string matches for the property IDs seem a little error-prone and not centralized. See the individual comment for a possible solution using the type system to help us here.

Eddykasp:
Error when clicking and dragging on diagram due to properties.get not being defined, requires some default values.

```
Uncaught TypeError: targetNode.parent.properties.get is not a function
    at KlighdInteractiveMouseListener.mouseDown (klighd-interactive-m…listener.js?d9c1:99)
    at eval (mouse-tool.js?d5a7:82)
    at Array.map (<anonymous>)
    at MouseTool.handleEvent (mouse-tool.js?d5a7:82)
    at MouseTool.mouseDown (mouse-tool.js?d5a7:119)
    at invokeHandler (eventlisteners.js?7eb2:4)
    at handleEvent (eventlisteners.js?7eb2:18)
    at SVGSVGElement.handler (eventlisteners.js?7eb2:23)
```

packages/klighd-interactive/src/constraint-classes.ts
```diff
@@ -52,10 +52,11 @@ export class KNode extends RectangularNode implements KGraphElement {
     areChildAreaChildrenRendered = false
     areNonChildAreaChildrenRendered = false
     hasFeature(feature: symbol): boolean {
-        return feature === selectFeature || (feature === moveFeature && (this.parent as KNode).properties.interactiveLayout)
+        return feature === selectFeature || (feature === moveFeature && (this.parent as KNode).properties['org.eclipse.elk.interactiveLayout'] as boolean)
     }
 
-    properties: NodeProperties
+    //properties: NodeProperties
```
soerendomroes, line 58: Remove this line

packages/klighd-interactive/src/layered/constraint-utils.ts
```diff
@@ -71,7 +71,7 @@ export function getLayerOfNode(node: KNode, nodes: KNode[], layers: Layer[], dir
 export function getActualLayer(node: KNode, nodes: KNode[], layerCandidate: number): number {
 
     // Examine all nodes that have a layer Id left or equal to the layerCandidate and that have a layerCons > their layerId
-    const layerConstraintLeftOfCandidate = nodes.filter(n => n.properties.layerId <= layerCandidate && n.properties.layerConstraint > n.properties.layerId)
+    const layerConstraintLeftOfCandidate = nodes.filter(n => n.properties['org.eclipse.elk.layered.layering.layerId'] as number <= layerCandidate && (n.properties['org.eclipse.elk.layered.layering.layerChoiceConstraint'] as number) > (n.properties['org.eclipse.elk.layered.layering.layerId'] as number))
```
soerendomroes, line 74: I suggest to introduce a linebreak

soerendomroes (changes requested):

packages/klighd-core/src/views-rendering.tsx
```diff
@@ -1058,9 +1062,9 @@ export function getKRendering(datas: KGraphData[], context: SKGraphModelRenderer
             continue
         if (data.type === K_RENDERING_REF) {
             if (context.kRenderingLibrary) {
-                const id = (data as KRenderingRef).renderingId
+                const id = (data as KRenderingRef).properties['klighd.lsp.rendering.id'] as string
                 for (const rendering of context.kRenderingLibrary.renderings) {
-                    if ((rendering as KRendering).renderingId === id) {
+                    if ((rendering as KRendering).properties['klighd.lsp.rendering.id'] as string === id) {
                         context.boundsMap = (data as KRenderingRef).calculatedBoundsMap
```
Eddykasp, line 1068: What is a case where this situation is reached? @NiklasRentzCAU

NiklasRentzCAU: The KGraphDiagramSynthesis uses a rendering library for all default renderings of nodes and edges, if activated in the synthesis options. You can try that out with some .kgt file

packages/klighd-interactive/src/layered/constraint-utils.ts
```diff
@@ -326,7 +326,7 @@ export function isLayerForbidden(node: KNode, layer: number): boolean {
 
     // check the connected nodes for layer constraints
     for (const node of connectedNodes) {
-        if (node.properties['org.eclipse.elk.layered.layering.layerId'] === layer && node.properties['org.eclipse.elk.layered.layering.layerChoiceConstraint'] !== -1) {
+        if (node.properties['org.eclipse.elk.layered.layering.layerId'] === layer && node.properties['org.eclipse.elk.layered.layering.layerChoiceConstraint'] !== -1 && node.properties['org.eclipse.elk.layered.layering.layerChoiceConstraint'] !== undefined) {
```
soerendomroes, line 329: I would suggest semantic line breaks here.

PR #65, opened 2022-03-25
Fix multiline label position calculation.

soerendomroes:
Since text is vertically centered one has to calculate the start y coordinate by taking the width of the lines up to the middle line into account.


NiklasRentzCAU (changes requested):
Text is not always centered vertically, see for example multiline texts as rendered in the code synthesis when for example translating SCCharts to C code. I guess this change would break any multiline text that is not centered vertically.

I guess the place to solve this problem would be `views-common.ts`  in the `findTextBoundsAndTransformationData` function: there the bounds are not adapted to the label text override property yet. This would then also work for any vertical alignment for texts.

PR #69, opened 2022-04-27
fix: use gradient vector instead of gradient angle for colors.

### LINKED ISSUE #38: Gradient in SVG support broken
It seems to be the case that the gradient in the SCCharts visualization is not correctly exported. At least, if the file is opened in inkscape this is not the case.

To check:

What happens if SCChart is exported to svg in KIELER and VSCode

Do they look the same?
How about opening them in inkscape
How about exporting them to pdf via inkscape cli?

NiklasRentzCAU:
Same behavior, but works around an issue in Inkscape when opening with that. Fixes #38

### DIFF
--- a/packages/klighd-core/src/views-styles.tsx
+++ b/packages/klighd-core/src/views-styles.tsx
@@ -54,8 +54,6 @@ export const K_VERTICAL_ALIGNMENT = 'KVerticalAlignmentImpl'
 
 // constants for string building
 const GRADIENT_UNIT_OBJECT_BOUNDING_BOX = 'objectBoundingBox'
-const GRADIENT_TRANSFORM_ROTATE_START = 'rotate('
-const GRADIENT_TRANSFORM_ROTATE_END = ')'
 
 const RGB_START = 'rgb('
 const RGB_END = ')'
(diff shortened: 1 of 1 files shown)


packages/klighd-core/src/views-styles.tsx
```diff
@@ -394,13 +392,64 @@ export function colorDefinition(colorId: string, start: ColorStyle, end: ColorSt
...
+        x = Math.tan(1/2 * Math.PI - angleRad)
+        y = 1
+    } else if (angleRad <= 5/4 * Math.PI) {
+        x = -1
+        y = Math.tan(2/2 * Math.PI - angleRad)
+    } else { // or: else if (angleRad <= 7/4 * Math.PI) {
+        x = - Math.tan(3/2 * Math.PI - angleRad)
+        y = -1
+    }
+
```
Eddykasp, line 421: shouldn't this all directly be solvable with `x = Math.cos(angleRad)` and `y = Math.sin(angleRad)`?

Eddykasp: Ok I understand, looks good to me then.

Eddykasp (approved):

PR #76, opened 2022-05-03
remove unused tooltip field

### LINKED ISSUE #73: Put Tooltip back into the property field
The "tooltip" property is still a direct field of the SKGraph elements and not put back into its properties field as initially in the KGraph structure. Supplements #63.
Also requires the same change in the KLighD Repository.

Eddykasp:
Resolves #73 

The tooltip is already stored in the properties, this PR removes the unused tooltip fields.

[Server-side PR](https://github.com/kieler/KLighD/pull/115)


NiklasRentzCAU (approved):
So the tooltips already worked with the new system with the properties and this field was now unused? I thought this would require more changes to make tooltips work in the new system again. But if that is not the case and no more changes are needed, LGTM.

PR #86, opened 2022-05-23
Code Review: tik/proxy-view

fight4day:
Draft of the proxy-view. This pull request is used as means of a code review.

https://github.com/kieler/KLighD/pull/129


package.json
```diff
@@ -19,7 +19,7 @@
         "lint": "lerna run lint --parallel",
         "prebuild": "yarn clean",
         "build": "lerna run build",
-        "watch": "lerna run watch --parallel",
+        "watch": "lerna run --scope @kieler/klighd-core --scope klighd-vscode watch --parallel",
```
Eddykasp, line 22: this shouldn't be in the final PR

packages/klighd-core/src/depth-map.ts
```diff
@@ -393,6 +394,7 @@ export class DepthMap {
      * @returns Boolean value indicating the visibility of the region in the current viewport. 
      */
     isInBounds(region: Region, viewport: Viewport): boolean {
+        // TODO: only partially working, e.g. region "too late" out of bounds, see motor-copy
```
Eddykasp, line 397: this is also just unfortunate naming here, because depthmap regions are not the same as sccharts regions

packages/klighd-core/src/options/general-panel.tsx
```diff
@@ -139,7 +139,8 @@ export class GeneralPanel extends SidebarPanel {
                 <div class-options__section="true">
                     <h5 class-options__heading="true">Render Options</h5>
                     {this.optionsRenderer.renderRenderOptions(
-                        this.renderOptionsRegistry.allRenderOptions
+                        this.renderOptionsRegistry.allRenderOptions,
+                        null
```
Eddykasp, line 143: what does null do here?

packages/klighd-core/src/proxy-view/proxy-view-actions.ts
```diff
@@ -0,0 +1,117 @@
...
+            // Register to receive updates on registry changes
+            if (!this.onChangeRegistered) {
+                // Make sure the rendering cache is cleared when the renderings change
+                this.synthesesRegistry.onChange(() => { this.proxyView.clearRenderings(); this.proxyView.clearPositions() });
+                this.optionsRegistry.onChange(() => this.proxyView.clearRenderings());
+                // Make sure to be notified when rendering options are changed
+                this.renderOptionsRegistry.onChange(() => this.proxyView.updateOptions(this.renderOptionsRegistry));
+                this.onChangeRegistered = true;
+            }
+        } else if (action.kind === SendModelContextAction.KIND && this.proxyView !== undefined) {
```
Eddykasp, line 86: for readability, this and the next case should have the same ordering i.e. either check action type first in both cases or check whether proxy view is defined first in both cases

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,742 @@
...
+    }
+
+    //////// Main methods ////////
+
+    // !!! TODO: might be a useful addition to save absolute coords in SKNode, not my task but also required here
+
+    /** Updates the proxy-view. */
+    update(model: SGraph, ctx: SKGraphModelRenderer): void {
+        if (!this.proxyViewEnabled) {
+            this.currHTMLRoot = this.patcher(this.currHTMLRoot, <div />);
```
Eddykasp, line 111: if the proxy view is disabled do you really still have to patch an empty div?

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,742 @@
...
+        if (!depthMap || !viewport) {
+            // Not yet initialized
+            return [];
+        }
+
+        //// Initial nodes ////
+        const { offScreenNodes, onScreenNodes } = this.getOffAndOnScreenNodes(root, ctx);
+
+        //// Apply filters ////
+        const filteredOffScreenNodes = this.applyFilters(offScreenNodes, onScreenNodes);
```
Eddykasp, line 153: this reads a little confusing, the on screen nodes are passed in as a parameter for the filters and the result is a subset of the original off screen nodes, correct?

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,742 @@
...
+
+        // Calculate size
+        const size = Math.min(canvasWidth, canvasHeight) * this.sizePercentage * 0.08;
```
Eddykasp, line 156: magic number

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,742 @@
...
+                    onScreenNodes.push(...offAndOnScreenNodes.onScreenNodes);
+                }
+            }
+        }
+
+        return { offScreenNodes: offScreenNodes, onScreenNodes: onScreenNodes };
+    }
+
+    /** Applies clustering to all `offScreenNodes` until there's no more overlap. Cluster-proxies are returned as VNodes. */
+    private applyClustering(offScreenNodes: { node: SKNode, transform: TransformAttributes }[],
```
Eddykasp, line 220: I think this clustering problem can be solved more efficiently using a sweep line approach, might be worth looking into

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,742 @@
...
+        const id = this.getProxyId(node.id);
+        let vnode = this.renderings.get(id);
+        if (!vnode) {
+            // Node hasn't been rendered yet (cache empty)
+
+            // Clone the node
+            const clone: SKNode = Object.create(node);
+            // Change its id for good measure
+            clone.id = id;
+            // Clear children, proxies don't show nested nodes (but keep Labels)
```
Eddykasp, line 340: what about ports?

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,742 @@
...
+            // Check if synthesis has specified a proxy rendering
+            if (this.useSynthesisProxyRendering && node.properties && node.properties[ProxyView.PROXY_RENDERING_PROPERTY]) {
+                // Proxy rendering available
+                console.log("node"); // FIXME:
+                console.log(node.properties[ProxyView.PROXY_RENDERING_PROPERTY]); // FIXME:
+                console.log("proxyRendering"); // FIXME:
+                console.log(node.properties[ProxyView.PROXY_RENDERING_PROPERTY + "2"]); // FIXME:
+                clone.data = node.properties[ProxyView.PROXY_RENDERING_PROPERTY] as KGraphData[];
+            } else {
+                // Fallback, use mock
```
Eddykasp, line 354: careful with overusing the word mock here, in this case it is simply the client-side default/fallback handling, mock would either refer to the mockup (design) or a mock object (standin during development of unimplemented interfaces)

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,742 @@
...
+                // TODO: further specify what to change for the mock?
+            }
+
+            vnode = ctx.renderProxy(clone);
+
+            // OLD: code to make a proxy non-click-through
```
Eddykasp, line 360: is this done somewhere else now and no longer necessary?

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,742 @@
...
+        const offsetY = 0.5 * (node.bounds.height * zoom - proxyHeight);
+        let x = (pos.x - scroll.x) * zoom + offsetX;
+        let y = (pos.y - scroll.y) * zoom + offsetY;
+
+        // Cap proxy at canvas border
+        x = Math.max(0, Math.min(canvasWidth - proxyWidth, x));
+        y = Math.max(0, Math.min(canvasHeight - proxyHeight, y));
+
+        // Make sure the proxies aren't rendered behind the sidebar buttons at the top right
+        /* Don't need to check for the opened sidebar since it closes as soon as the diagram is moved
```
Eddykasp, line 458: but if there is a proxy where the sidebar opens, should it not then adjust its position?

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,742 @@
...
+        // x /= scale;
+        // y /= scale;
+
+        return { x: x, y: y, scale: scale, width: proxyWidth, height: proxyHeight };
+    }
+
+    /** Recursively calculates the positions of this node and all of its predecessors and stores them in {@link positions}. */
+    private getPosition(node: SKNode): Point {
+        if (!node) {
+            return { x: 0, y: 0 };
```
Eddykasp, line 484: if something undefined is passed here, wouldn't it make more sense to return undefined? this could promote misusing the function

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,742 @@
...
+                                        "text": "${numProxies}"
+                                    }
+                                ]
+                            }
+                        ],
+                        "key": "$${id}2"
+                    }
+                ],
+                "key": "$${id}"
+            }`);
```
Eddykasp, line 718: is json.parse the best solution here?

packages/klighd-core/src/uiextension-minimal-example/example-actions.ts
```diff
@@ -0,0 +1,32 @@
...
+export type ShowExampleAction = SetUIExtensionVisibilityAction
+
+export namespace ShowExampleAction {
+    export function create(): ShowExampleAction {
+        return SetUIExtensionVisibilityAction.create({
+            extensionId: Example.ID,
+            visible: true,
+        })
+    }
+}
```
Eddykasp, line 32: This example shouldn't be in the final PR, whole folder not just this file

packages/klighd-core/src/views-rendering.tsx
```diff
@@ -527,7 +527,7 @@ export function renderKText(rendering: KText,
 export function renderChildRenderings(parentRendering: KContainerRendering, parentElement: SKGraphElement, propagatedStyles: KStyles,
     context: SKGraphModelRenderer, childOfNodeTitle?: boolean): (VNode | undefined)[] {
     // children only should be rendered if the parentElement is not a shadow
-    if (!(parentElement instanceof SKNode) || !parentElement.shadow) {
+    if (parentRendering.children && (!(parentElement instanceof SKNode) || !parentElement.shadow)) {
```
Eddykasp, line 530: what makes this change necessary?

packages/klighd-core/src/depth-map.ts
fight4day: It was an old TODO, I removed it. Thanks for remark

packages/klighd-core/src/proxy-view/proxy-view.tsx
fight4day: Thanks, I now prevent excessive patching by saving the previous value of this.proxyViewEnabled

packages/klighd-core/src/proxy-view/proxy-view.tsx
fight4day: Some changes to the user API et voilà, magic number no more

packages/klighd-core/src/proxy-view/proxy-view.tsx
fight4day: After giving it some thought I decided to also remove ports

packages/klighd-core/src/proxy-view/proxy-view.tsx
fight4day: As far as I can tell yes, I've outsourced the method as discussed

packages/klighd-core/src/proxy-view/proxy-view.tsx
fight4day: Judging by what I've read so far, the approach can be fit to work for clustering but requires complex data structures to actually reach the O(n*log(n)) runtime. With what would be implementable without going too deep down that rabbit hole, I may be able to improve the average case - though worst case would still be O(n²). I'll look at that when I have some spare time. Thanks for the suggestion!

Eddykasp (commented):
looks good to me so far

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,650 @@
...
+            document.getElementById(`keith-diagram_sprotty_${node.key?.toString()}`)?.setAttribute("transform", transformString);
+        }
+    }
+
+    /**
+     * Checks if the given bounds overlap.
+     * 
+     * @returns `true` if there is overlap.
+     */
+    private checkOverlap(bounds1: TransformAttributes, bounds2: TransformAttributes): boolean {
```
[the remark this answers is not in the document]
fight4day: Good suggestion!

packages/klighd-core/src/proxy-view/proxy-view-actions.ts
```diff
@@ -0,0 +1,119 @@
...
+ * Handles {@link SendProxyViewAction}s to get the {@link ProxyView} instance
+ * aswell as {@link SendModelContextAction}s to redirect the content to the {@link ProxyView}.
+ */
+@injectable()
+export class ProxyViewActionHandler implements IActionHandler, IActionHandlerInitializer {
+    private proxyView: ProxyView;
+    @inject(DISymbol.SynthesesRegistry) private synthesesRegistry: SynthesesRegistry;
+    @inject(DISymbol.RenderOptionsRegistry) private renderOptionsRegistry: RenderOptionsRegistry;
+    @inject(DISymbol.OptionsRegistry) private optionsRegistry: OptionsRegistry;
+    /** Whether the proxy-view was registered in the registries' onchange() method. Prevents registering multiple times. */
```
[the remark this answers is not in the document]
NiklasRentzCAU: Why can't it be registered once during initialization to avoid this altogether?

packages/klighd-core/src/proxy-view/proxy-view-cluster.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const squareSize = size * 0.9;
+    const squareOffset = size * 0.1;
+    const textSize = size * 0.5;
+    return JSON.parse(
+        `{
+            "sel": "g",
+            "data": {
+                "ns": "http://www.w3.org/2000/svg",
+                "attrs": {
+                    "id": "keith-diagram_sprotty_$${id}",
```
NiklasRentzCAU, line 14: Does this guarantee unique IDs? Otherwise some $-separated proxy view prefix would be good

packages/klighd-core/src/proxy-view/proxy-view-module.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { ContainerModule } from "inversify";
+import { TYPES } from "sprotty";
+import { DISymbol } from "../di.symbols";
+import { ProxyView } from "./proxy-view";
+import { ProxyViewActionHandler } from "./proxy-view-actions";
+
+export const proxyViewModule = new ContainerModule((bind) => {
+    // The class needs to be bound to itself and
+    // we have to let sprotty know our service is an UIExtension
```
NiklasRentzCAU, line 26: Sprotty instead of sprotty
It's 'a' UIExtension, not 'an'

packages/klighd-core/src/proxy-view/proxy-view-cluster.tsx
```diff
@@ -0,0 +1,150 @@
...
+                                            "dominant-baseline": "middle",
+                                            "font-family": "overpass, sans-serif",
+                                            "font-size": "${textSize}",
+                                            "opacity": 1
+                                        },
+                                        "attrs": {
+                                            "x": 0,
+                                            "xml:space": "preserve",
+                                            "y": 0,
+                                            "lengthAdjust": "spacingAndGlyphs"
```
NiklasRentzCAU, line 137: This attribute does nothing if you do not set a predefined text width. Can be removed.

packages/klighd-core/src/proxy-view/proxy-view-cluster.tsx
```diff
@@ -0,0 +1,150 @@
+import { VNode } from "snabbdom";
+
+/** Returns the rendering of clusters. */
+export function getClusterRendering(id: string, numProxies: number, size: number, x: number, y: number): VNode {
+    const squareSize = size * 0.9;
+    const squareOffset = size * 0.1;
+    const textSize = size * 0.5;
+    return JSON.parse(
```
NiklasRentzCAU, line 8: I still do not like this very much. What attribute prevented you from simply using jsx here?

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,678 @@
...
+    ProxyViewCapProxyToParent, ProxyViewClusteringCascading, ProxyViewClusteringEnabled, ProxyViewClusteringSweepLine, ProxyViewEnabled,
+    ProxyViewFilterUnconnected, ProxyViewSize, ProxyViewUseSynthesisProxyRendering, RenderOptionsRegistry
+} from "../options/render-options-registry";
+import { SKGraphModelRenderer } from "../skgraph-model-renderer";
+import { SKEdge, SKNode, SKPort } from "../skgraph-models";
+import { SendProxyViewAction, ShowProxyViewAction, TransformAttributes } from "./proxy-view-actions";
+import { getClusterRendering } from "./proxy-view-cluster";
+
+@injectable()
+export class ProxyView extends AbstractUIExtension {
```
NiklasRentzCAU, line 35: class doc comment missing

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,678 @@
...
+            return node instanceof SKNode && node.id.charAt(node.id.lastIndexOf("$") - 1) !== "$";
+        }
+    }
+
+    /** Updates the proxy-view options specified in the {@link RenderOptionsRegistry}. */
+    updateOptions(renderOptionsRegistry: RenderOptionsRegistry): void {
+        this.prevProxyViewEnabled = this.proxyViewEnabled;
+        this.proxyViewEnabled = renderOptionsRegistry.getValue(ProxyViewEnabled);
+
+        const toPercent = 0.01;
```
NiklasRentzCAU, line 469: fromPercent, as you convert the value from a percentage to a real numbered value?

packages/klighd-core/src/skgraph-model-renderer.ts
```diff
@@ -77,4 +78,12 @@ export class SKGraphModelRenderer extends ModelRenderer {
             })
             .filter(vnode => vnode !== undefined) as VNode[]
     }
+
+    /** Renders a node as a proxy. */
+    renderProxy(node: SKNode): VNode | undefined {
+        this.forceRendering = true;
+        const vnode = super.renderElement(node);
+        this.forceRendering = false;
```
NiklasRentzCAU, line 86: What if something else set the forceRendering to true and it should not be reset? This method could just restore the old state of forceRendering instead of setting it to false.

packages/klighd-core/src/views-rendering.tsx
```diff
@@ -527,7 +527,7 @@ export function renderKText(rendering: KText,
 export function renderChildRenderings(parentRendering: KContainerRendering, parentElement: SKGraphElement, propagatedStyles: KStyles,
     context: SKGraphModelRenderer, childOfNodeTitle?: boolean): (VNode | undefined)[] {
     // children only should be rendered if the parentElement is not a shadow
-    if (!(parentElement instanceof SKNode) || !parentElement.shadow) {
+    if (parentRendering.children && (!(parentElement instanceof SKNode) || !parentElement.shadow)) { // FIXME: is checking parentRendering.children actually necessary?
```
NiklasRentzCAU, line 530: is it?

packages/klighd-core/src/views.tsx
```diff
@@ -39,10 +40,16 @@ export class SKGraphView implements IView {
...
     @inject(DISymbol.RenderOptionsRegistry) renderOptionsRegistry: RenderOptionsRegistry
+    @inject(TYPES.IActionDispatcher) private actionDispatcher: IActionDispatcher;
 
     render(model: Readonly<SGraph>, context: RenderingContext): VNode {
         const ctx = context as SKGraphModelRenderer
-        ctx.renderingDefs = new Map
+        this.actionDispatcher.dispatch(SendModelContextAction.create(model, ctx))
+
+        if (!ctx.renderingDefs) {
+            // Make sure not to create a new map all the time
```
NiklasRentzCAU, line 50: Could this be adding the font defs, even if they already exist though?

packages/klighd-core/src/proxy-view/proxy-view-actions.ts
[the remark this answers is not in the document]
fight4day: Since there are multiple registries plural should be correct.
The registries cannot be injected in the ProxyView since it results in a circular dependency - therefore it can't be done in the ProxyView's init() method. Regarding an init() method in the ProxyViewActionHandler, the proxyView field is most likely undefined when init() is called. Therefore doing it this way is the 'safe' way, since you only register once the ProxyView is sure to be instantiated.
Although the SendProxyViewAction should only be sent once anyway, but well, better to be safe than sorry

packages/klighd-core/src/proxy-view/proxy-view-cluster.tsx
fight4day: I'm pretty sure it should be unique already since there's no $root and when it's called by the clustering the given id definitely makes it unique. But a $-prefix doesn't hurt, so I added it

packages/klighd-core/src/proxy-view/proxy-view-cluster.tsx
fight4day: `width` and `height` only have getters and couldn't be set easily. This seemed like the simplest way to get around this

packages/klighd-core/src/proxy-view/proxy-view.tsx
fight4day: Or you convert an integer to an actual percentage

packages/klighd-core/src/skgraph-model-renderer.ts
fight4day: Good remark, I didn't think about others potentially using the field I added aswell

packages/klighd-core/src/views-rendering.tsx
fight4day: I'll find out once I've finished the synthesis-related stuff

packages/klighd-core/src/views.tsx
fight4day: Since it's a map they would just be replaced as it was done previously

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -476,6 +465,13 @@ export class ProxyView extends AbstractUIExtension {
 
     //////// General helper methods ////////
 
+    /** Checks if `b1` is (partially) in `b2`. */
```
Eddykasp, line 468: I think the decoupling from the depth map is good. Especially since what you used from it was only this very specific functionality. I also think that this `isInBounds` method is more general than any specific feature and should reside in a more general place. There are a few different -util files, maybe there is something suitable there already.

fight4day: Good idea, I was hoping in the foreseeable futurue we could look at the methods together to see which ones could be moved to general util-functions (and which util-file to add them to)

packages/klighd-core/src/proxy-view/proxy-view.tsx
NiklasRentzCAU: The integer (e.g. 5 (%)) is the percentage, the real-numbered value 0.05 is not a percentage

packages/klighd-core/src/proxy-view/proxy-view-cluster.tsx
NiklasRentzCAU: Compare to the `renderSingleSVGRect` method in views-rendering: There the width and height are just set as attributes using jsx, so that alone cannot be the reason

fight4day:
> Could it be that the id change is persistent, on the SKEdge, between renderings and that both the edge and its proxy are rendered?
> In the next rendering the original edge would have the proxy's id and as getProxId does not add the proxy suffix when already present its not added again for the proxy resulting in them having the same id.

Thanks for the input! I get what you mean and looked at it but believe that's not what's causing the bug. If that was the case, the error would appear anytime the cloned SKEdge is rendered for the second time and as it stands right now, the error is logged only in some cases for large diagrams.
Also, it would be weird for the same thing to work with SKNodes but not SKEdges, but then again, the handling might be different in unexpected ways :v

At the moment I think the edge proxies are causing it themselves. Somehow they seem to appear on the document twice with the same ids..? I'll definitely keep it in mind though

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -348,459 +530,512 @@ export class ProxyView extends AbstractUIExtension {
...
+        return vnode;
+    }
 
-        const useSynthesisProxyRendering = renderOptionsRegistry.getValue(ProxyViewUseSynthesisProxyRendering);
-        if (this.useSynthesisProxyRendering !== useSynthesisProxyRendering) {
-            // Make sure not to use the wrong renderings if changed
-            this.clearRenderings();
-        }
-        this.useSynthesisProxyRendering = useSynthesisProxyRendering;
+    //////// General helper methods ////////
```
Eddykasp, line 751: are these not also methods for the utils?

packages/klighd-core/src/options/render-options-registry.ts
```diff
@@ -19,7 +19,8 @@ import { inject, injectable, postConstruct } from "inversify";
...
-import { ProxyViewCapProxyToParent, ProxyViewCategory, ProxyViewClusteringCascading, ProxyViewClusteringEnabled, ProxyViewClusteringSweepLine, ProxyViewDebugCategory, ProxyViewEnabled, ProxyViewFilterDistant, ProxyViewFilterUnconnected, ProxyViewHighlightSelected, ProxyViewSize, ProxyViewUseSynthesisProxyRendering } from "../proxy-view/proxy-view-options";
+import { ProxyViewFilterCategory, ProxyViewFilterDistant, ProxyViewFilterUnconnectedToOnScreen, ProxyViewFilterUnconnectedToSelected } from "../proxy-view/filters/proxy-view-filter-options";
+import { ProxyViewCapProxyToParent, ProxyViewCapScaleToOne, ProxyViewCategory, ProxyViewClusteringCascading, ProxyViewClusteringEnabled, ProxyViewClusteringSweepLine, ProxyViewClusterTransparent, ProxyViewDebugCategory, ProxyViewActionsEnabled, ProxyViewEnabled, ProxyViewHighlightSelected, ProxyViewOpacityByDistance, ProxyViewOpacityBySelected, ProxyViewSize, ProxyViewStackingOrderByDistance, ProxyViewUsePositionsCache, ProxyViewUseSynthesisProxyRendering, ProxyViewDrawStraightEdges, ProxyViewUseDetailLevel, ProxyViewStackingOrderByOpacity, ProxyViewStackingOrderBySelected, ProxyViewTitleScaling } from "../proxy-view/proxy-view-options";
```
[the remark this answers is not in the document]
fight4day: I'd love to just do import * as both of these files contain nothing but RenderOptions. Unfortunately, I don't think that's possible

packages/klighd-core/src/proxy-view/proxy-view-options.ts
```diff
@@ -17,10 +17,31 @@
...
+ * Size of Proxies?
+ *   range
+ * Decrease Proxy Clutter
+ *   Off
+ *   Clustering
+ *   Stacking Order + Transparency by Distance
+ * Enable Actions
+ *   true/false
+ * Cap Proxy to parent
+ *   true/false
```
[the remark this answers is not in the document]
fight4day: I believe this to be very dependant on the diagram type, e.g. diagrams that allow for direct connections between hierarchy levels (such as deployment diagrams). Potentially even something the synthesis may also want to have a word in, but that sort of leads back to semantic filtering

packages/klighd-core/src/proxy-view/proxy-view-options.ts
```diff
@@ -94,47 +160,27 @@ export class ProxyViewCapProxyToParent implements RenderOption {
     currentValue = ProxyViewCapProxyToParent.DEFAULT;
 }
 
-/** Whether proxies should be filtered by removing unconnected nodes. */
-export class ProxyViewFilterUnconnected implements RenderOption {
-    static readonly ID: string = 'proxy-view-filter-unconnected';
-    static readonly NAME: string = 'Filter Unconnected Nodes';
-    static readonly DESCRIPTION: string = 'Whether proxies should be filtered by removing unconnected nodes.';
+/** Whether to use title scaling if smart zoom is enabled. */
```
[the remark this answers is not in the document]
fight4day: I don't know about that one, I could imagine wanting to have smart zoom turned on solely for performance reasons but not wanting the title scaling enabled. If time permits, we could bring it up in an upcoming K-Meeting

packages/klighd-core/src/proxy-view/proxy-view-options.ts
```diff
@@ -176,11 +238,107 @@ export class ProxyViewUseSynthesisProxyRendering implements RenderOption {
...
+    readonly name: string = ProxyViewCapScaleToOne.NAME;
+    readonly type: TransformationOptionType = TransformationOptionType.CHECK;
+    readonly initialValue: boolean = ProxyViewCapScaleToOne.DEFAULT;
+    readonly description: string = ProxyViewCapScaleToOne.DESCRIPTION;
+    readonly renderCategory: RenderOption = ProxyViewDebugCategory.INSTANCE;
+    readonly debug: boolean = true;
+    currentValue = ProxyViewCapScaleToOne.DEFAULT;
+}
+
+/** Whether clusters should be transparent according to their contained proxies. */
```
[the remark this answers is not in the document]
fight4day: Yes, which can potentially lead to some unintuitive behavior

packages/klighd-core/src/proxy-view/proxy-view.tsx
fight4day: All of these are either proxy-view specific or cannot be outsourced any more, for example because of caches

Eddykasp: ok

packages/klighd-core/src/proxy-view/proxy-view-options.ts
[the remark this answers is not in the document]
fight4day: Nvm, I gave it some more thought and agree that this should be the standard for any diagram, as only the outermost node is represented as a proxy. I'll move it to debug

packages/klighd-core/src/proxy-view/proxy-view-options.ts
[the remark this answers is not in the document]
Eddykasp: yes, definitely something worth discussing

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -647,22 +724,241 @@ export class ProxyView extends AbstractUIExtension {
...
+                }
+                prevPoint = p;
+            }
+
+            if (canvasEdgeIntersections.length < 2) {
+                // Not enoug intersections to form a connection
+                continue;
+            }
+
+            // Make sure to only connect on-off-on intersections, not off-on-off
```
Eddykasp, line 879: not clear what on-off-on means

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -791,24 +1091,57 @@ export class ProxyView extends AbstractUIExtension {
...
             const translatedParent = this.getTranslatedNodeBounds(node.parent as SKNode, canvas);
-            x = Math.max(translatedParent.x, Math.min(translatedParent.x + translatedParent.width - proxyWidth, x));
-            y = Math.max(translatedParent.y, Math.min(translatedParent.y + translatedParent.height - proxyHeight, y));
+            x = capNumber(x, translatedParent.x, translatedParent.x + translatedParent.width - proxyWidth);
+            y = capNumber(y, translatedParent.y, translatedParent.y + translatedParent.height - proxyHeight);
         }
 
         return { x, y, scale, width: proxyWidth, height: proxyHeight };
+
+        // TODO: RF
```
Eddykasp, line 1104: is this still TODO/relevant?

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -891,35 +1224,115 @@ export class ProxyView extends AbstractUIExtension {
...
+            } else if (ctx.decorationMap && ctx.decorationMap[id]) {
                 // Arrow head was in rendering refs
                 props["klighd.lsp.calculated.decoration"] = { ...ctx.decorationMap[id], origin: target, rotation: angle };
             } else {
                 // Better not to show arrow head as it would be floating around somewhere
                 return [];
             }
+        } else if (clone.type === K_SPLINE) {
+            // TODO: spline type, for now just change type
+            K_POLYLINE
```
Eddykasp, line 1253: so now you're not changing it though, this line does nothing or am I missing something?

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -891,35 +1224,115 @@ export class ProxyView extends AbstractUIExtension {
...
+            return edgeData;
+        }
+
+        const res = [];
+        const clone = { ...data } as any;
+        const props = { ...clone.properties };
+        const styles = [...clone.styles];
+        clone.properties = props;
+        clone.styles = styles;
+        // OLD: so far this seems to only be used for "DefaultEdgeRendering" and doesn't seem to break anything if set
```
Eddykasp, line 1282: can these OLD comments be removed?

packages/klighd-core/src/proxy-view/proxy-view-util.ts
```diff
@@ -37,10 +37,374 @@ export interface TransformAttributes extends Bounds {
...
+        // Cap bounds at canvas border
+        let x = capNumber(bounds.x, canvas.x, canvas.x + canvas.width - bounds.width);
+        const y = capNumber(bounds.y, canvas.y, canvas.y + canvas.height - bounds.height);
+
+        // Make sure the proxies aren't rendered behind the sidebar buttons at the top right
+        // Don't reposition proxies with an open sidebar since it closes as soon as the diagram is moved (onMouseDown)
+        const rect = document.querySelector(".sidebar__toggle-container")?.getBoundingClientRect();
+        const isSidebarOpen = document.querySelector(".sidebar--open");
+        if (!isSidebarOpen && rect) {
+            const rect2 = rect;//Canvas.translateToLRF(rect, canvas); // TODO: RF
```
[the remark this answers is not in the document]
fight4day: I was testing around with reference frames, but came to the conclusion that changing the reference frame won't help for this function

packages/klighd-core/src/proxy-view/proxy-view-util.ts
```diff
@@ -79,120 +445,170 @@ export function checkOverlap(b1: Bounds, b2: Bounds): boolean {
...
+            // Intersection at y, find x
+            const topOrBottom = p1.y < b.y ? b.y : b.y + b.height;
+
+            // Scalar of line equation, must be in [0,1] as to not be before p2 or after p1, could be ±inf
+            const scalar = capNumber((topOrBottom - p2.y) / (p1.y - p2.y), 0, 1);
+
+            // Intersection point
+            const intersectX = p2.x + scalar * (p1.x - p2.x);
+            return { x: intersectX, y: topOrBottom };
+        }
```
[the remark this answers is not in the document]
fight4day: How?

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -24,23 +24,37 @@ import { Action, angleOfPoint, Bounds, CenterAction, FitToScreenAction, Point }
...
+    static readonly RENDER_NODE_AS_PROXY_PROPERTY = "de.cau.cs.kieler.klighd.proxy-view.renderNodeAsProxy";
+    /**
+     * ID used for specifying depth of going into hierarchical off-screen nodes.
+     * `0` indicates default behavior, showing only the outermost node as a proxy.
+     * A value `x>0` indicates showing proxies up to x layers deep inside a hierarchical node.
+     * A value `x<0` indicates always showing proxies for all layers.
+     */
+    static readonly HIERARCHICAL_OFF_SCREEN_DEPTH = "de.cau.cs.kieler.klighd.proxy-view.hierarchicalOffScreenDepth";
     /** Number indicating at what distance a node is close. */
     static readonly DISTANCE_CLOSE = 300;
```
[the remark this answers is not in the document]
fight4day: Back then we talked about this quite a bit. Leaving this up to the synthesis might be a good idea but potentially requires some remodelling as the value is used outside of this class aswell. For now, I'd say it's low priority but I'll look at it once I have spare time

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -217,66 +230,92 @@ export class ProxyView extends AbstractUIExtension {
...
                 this.currProxies.push({ proxy, transform });
             }
         }
 
+        const edgeProxies = [];
+        // Edges
+        const edges = ([] as { edge: SKEdge, transform: TransformAttributes }[]).concat(
+            connectors.overlayEdges, // Start with overlays to not have overlays over proxy edges
+            connectors.proxyEdges,
+            routedEdges.overlayEdges, // But routing overlays should still be over connectors
```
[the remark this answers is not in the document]
fight4day: I think they're convenient as you know exactly what part of the code the comment refers to. In this case, I would otherwise need to say which edges are routing overlays. What do you think of
```
const edges = ([] as { edge: SKEdge, transform: TransformAttributes }[]).concat(
    // Start with overlays to not have overlays over proxy edges
    connectors.overlayEdges,
    connectors.proxyEdges,
    // But routing overlays should still be over connectors
    routedEdges.overlayEdges, 
    ...
```
as a middle ground?

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -550,23 +602,22 @@ export class ProxyView extends AbstractUIExtension {
...
-                if (x > 0 && x < canvas.width - size && (y < canvas.height - size || y > 0)) {
-                    y = y > (canvas.height - size) / 2 ? canvas.height - size : 0;
+                } else if (x > canvas.x && x < canvas.x + canvas.width - size && (y > canvas.y || y < canvas.y + canvas.height - size)) {
+                    y = y > (canvas.height - size) / 2 ? canvas.y + canvas.height - size : canvas.y;
                     floating = true;
                 }
                 if (floating) {
                     // Readjust if it was previously floating
-                    ({ x, y } = capToCanvas({ x, y, width: size, height: size }, canvas));
+                    ({ x, y } = Canvas.capToCanvas({ x, y, width: size, height: size }, canvas));
```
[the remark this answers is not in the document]
fight4day: I could get rid of the if (floating) but then I'd always call a method that does some work and potentially ends up not changing anything. This is really only relevant for cascading clustering, which won't be used outside of debugging

packages/klighd-core/src/proxy-view/proxy-view.tsx
fight4day: I left it to not have an unused import. Can be removed since I won't change it

packages/klighd-core/src/proxy-view/proxy-view-util.ts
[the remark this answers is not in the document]
Eddykasp: I didn't look in detail, but there are some things you compute in several cases such as topOrBottom this stuff could be done outside the case distinctions if that doesn't work nicely maybe just small helper function.

packages/klighd-core/src/proxy-view/proxy-view.tsx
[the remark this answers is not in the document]
Eddykasp: my issue is that I think this muddles readability a bit, because it makes these look like distinct lines of code which they aren't, but maybe it's not really an issue. I'm open for other opinions on the matter

packages/klighd-core/src/proxy-view/filters/proxy-view-filter-options.ts
```diff
@@ -0,0 +1,109 @@
...
+ */
+
+import { RenderOption, TransformationOptionType } from "../../options/option-models";
+import { ProxyViewCategory } from "../proxy-view-options";
+
+/**
+ * Whether proxy-view debug filter should be hidden from the sidebar.
+ * `true` hides all debug filter, `false` shows all debug filter.
+ */
+const hideProxyViewDebugFilter = true;
```
Eddykasp, line 25: naming convention should be caps?

packages/klighd-core/src/proxy-view/filters/proxy-view-semantic-filters.ts
```diff
@@ -0,0 +1,172 @@
...
+                ProxySemanticFilterHandler,
+                ({ node }: ProxyFilterArgs) => !this.semanticFilterOptionValues[i] || filter.filterFun(node),
+                `${filter.name ?? "unknown"}-${i}`)
+        ));
+        // Also map to RenderOption
+        this.semanticFilterOptions = filters.map((filter, i) => (
+            class ProxyViewSemanticFilter extends ProxyViewAbstractSemanticFilter implements RenderOption {
+                static readonly ID: string = `proxy-view-semantic-filter-${filter.name ?? "unknown"}-${i}`;
+                static readonly NAME: string = filter.name ?? "Unknown Filter";
+                static readonly DEFAULT: boolean = !!filter.defaultValue;
```
Eddykasp, line 104: why double !! negation

packages/klighd-core/src/proxy-view/proxy-view-actions.ts
```diff
@@ -0,0 +1,169 @@
...
+    export function create(): ShowProxyViewAction {
+        return SetUIExtensionVisibilityAction.create({
+            extensionId: ProxyView.ID,
+            visible: true,
+        })
+    }
+}
+
+/** An action containing the {@link ProxyView}. */
+// Sent from the proxy-view to the action handler to avoid stackoverflows
```
Eddykasp, line 45: tidy up docstring formatting

packages/klighd-core/src/proxy-view/proxy-view-util.ts
```diff
@@ -0,0 +1,1001 @@
...
+    /**
+     * Returns the bounds translated from the LRF to the CRF.
+     * E.g. calculates its position & width/height according to scroll and zoom.
+     * Inverse to {@link translateToLRF()}.
+     * @param bpd The bounds/point/dimension in the LRF.
+     * @param canvas The canvas.
+     * @returns The bounds translated to the CRF.
+     */
+    export function translateToCRF(bpd: Bounds | Point | Dimension, canvas: Canvas): Bounds {
+        const b = asBounds(bpd);
```
Eddykasp, line 79: const naming conventions and maybe more readable names

packages/klighd-core/src/proxy-view/proxy-view-util.ts
```diff
@@ -0,0 +1,1001 @@
...
+    /**
+     * Returns the given bounds capped to the canvas border w.r.t. the sidebar if enabled.
+     * Note that `bp` and `canvas` need to be in CRF.
+     * Also, `bp` has to contain the absolute position (not relative to parent).
+     * @param bp The bounds/point to cap to the canvas border, absolute.
+     * @param canvas The canvas.
+     * @param capToSidebar Whether the bounds should also be capped to the sidebar.
+     * @returns The given bounds capped to the canvas border w.r.t. the sidebar if enabled.
+     */
+    export function capToCanvas(bp: Bounds | Point, canvas: Bounds, capToSidebar = true): Bounds {
```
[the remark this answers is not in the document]
Eddykasp: and update docstring to reflect renaming

packages/klighd-core/src/proxy-view/proxy-view-util.ts
```diff
@@ -0,0 +1,1001 @@
...
+ */
+export function checkOverlap(b1: Bounds, b2: Bounds): boolean {
+    return isInBounds(b1, b2) || isInBounds(b2, b1);
+}
+
+/**
+ * Returns `bpd` if given bounds, otherwise fills the empty attributes with zeros.
+ * @param bpd The bounds/point/dimension.
+ */
+export function asBounds(bpd: Bounds | Point | Dimension): Bounds {
```
Eddykasp, line 464: rename

packages/klighd-core/src/proxy-view/proxy-view-util.ts
```diff
@@ -0,0 +1,1001 @@
...
+ * @param n The number to cap.
+ * @param min The lower bound of the range.
+ * @param max The upper bound of the range.
+ * @returns `n` capped to the given range. If `min > max`, the two are swapped.
+ */
+export function capNumber(n: number, min: number, max: number): number {
+    if (min > max) {
+        const temp = min;
+        min = max;
+        max = temp;
```
Eddykasp, line 479: instead: ```return capNumber(n, max, min)```

packages/klighd-core/src/proxy-view/proxy-view-util.ts
```diff
@@ -0,0 +1,1001 @@
...
+        while (next) {
+            allSelectedElements.push(next);
+            queue.push(...(next.children as unknown as SKGraphElement[]));
+            next = queue.pop();
+        }
+        this.clearCaches(allSelectedElements);
+        this.selectedElements = allSelectedElements;
+    }
+
+    //// Util methods ////
```
Eddykasp, line 894: and the above are not util methods?

packages/klighd-core/src/skgraph-model-renderer.ts
```diff
@@ -77,4 +78,13 @@ export class SKGraphModelRenderer extends ModelRenderer {
...
+
+    /** Renders an element forcefully, i.e. independant of the depthMap. */
+    forceRenderElement(element: SKGraphElement): VNode | undefined {
+        const prevForceRendering = this.forceRendering;
+        this.forceRendering = true;
+        const vnode = super.renderElement(element);
+        this.forceRendering = prevForceRendering;
+        return vnode;
+    }
 }
```
Eddykasp, line 90: no new line at end of file

packages/klighd-core/src/filtering/semantic-filtering-util.ts
```diff
@@ -0,0 +1,591 @@
...
+export function getFilters(graph: SKGraphElement): Array<Filter> {
+    const filters: Array<Filter> = [];
+    if (graph.properties['de.cau.cs.kieler.klighd.semanticFilter.rules'] !== undefined) {
+        (graph.properties['de.cau.cs.kieler.klighd.semanticFilter.rules'] as Array<SemanticFilterRule>)
+            .forEach((rule) => {
+                filters.push(createFilter(rule));
+            });
+    }
+    return filters;
+}
```
Eddykasp, line 591: no newline at end of file

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,1530 @@
...
+     * Note that an off-screen node's children aren't included in the list, e.g. only outer-most off-screen nodes are returned.
+     */
+    private getOffAndOnScreenNodes(currRoot: SKNode, canvasLRF: Canvas, depth: number, ctx: SKGraphModelRenderer)
+        : { offScreenNodes: SKNode[], onScreenNodes: SKNode[] } {
+        // For each node check if it's off-screen
+        const offScreenNodes = [];
+        const onScreenNodes = [];
+        for (const node of currRoot.children) {
+            if (node instanceof SKNode) {
+                const b = this.getAbsoluteBounds(node);
```
Eddykasp, line 359: rename b

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,1530 @@
...
+        if (!this.clusteringEnabled) {
+            // Makes no sense to order when clustering is enabled since proxies cannot be stacked
+
+            /*
+            Order these stacking order criteria such that each criterion is more important the previous one,
+            i.e. the least important criterion is at the start and the most important one is at the end
+            */
+            if (this.stackingOrderByDistance) {
+                // Distant nodes at start, close nodes at end
+                res.sort((n1, n2) => this.getNodeDistanceToCanvas(n2, canvasLRF) - this.getNodeDistanceToCanvas(n1, canvasLRF));
```
Eddykasp, line 500: consider renaming n1, n2

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,1530 @@
...
+        }
+
+        return { edge: clone, transform: Bounds.EMPTY };
+    }
+
+    /** Returns an edge that can be overlayed over the given `edge` to simulate a fade-out effect. */
+    private getOverlayEdge(edge: SKEdge, canvas: Canvas, ctx: SKGraphModelRenderer): { edge: SKEdge, transform: TransformAttributes } {
+        // Color/opacity for fade out effect
+        const color = { red: 255, green: 255, blue: 255 };
+        const opacity = 0.8;
```
Eddykasp, line 914: magic values, should be defined as consts somewhere accessibly

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,1530 @@
...
+        this.clickThrough = false;
+        this.currProxies.forEach(({ proxy }) => updateClickThrough(proxy, !this.interactiveProxiesEnabled));
+    }
+
+    /** Updates the proxy-view options specified in the {@link RenderOptionsRegistry}. */
+    updateOptions(renderOptionsRegistry: RenderOptionsRegistry): void {
+        this.prevProxyViewEnabled = this.proxyViewEnabled;
+        this.proxyViewEnabled = renderOptionsRegistry.getValue(ProxyViewEnabled);
+
+        const fromPercent = 0.01;
```
Eddykasp, line 1411: magic number

Eddykasp (changes requested):
looking through the some of util functions I see some good example cases for unit tests

packages/klighd-core/src/views.tsx
NiklasRentzCAU: Only if the IDs are the same, I would take care with this if this can cause a new issue.

packages/klighd-core/src/proxy-view/proxy-view-util.ts
Eddykasp: the const is fine then, but i would still prefer the parameters to have better names. bpd for a parameter representing a bounds, point or dimension just doesn't really convey the intent

packages/klighd-core/src/proxy-view/proxy-view.tsx
Eddykasp: that's fair

packages/klighd-core/src/proxy-view/filters/proxy-view-filter-options.ts
```diff
@@ -0,0 +1,109 @@
...
+    readonly type: TransformationOptionType = TransformationOptionType.CATEGORY;
+    readonly initialValue: any;
+    readonly renderCategory: string = ProxyViewCategory.ID;
+    currentValue: any;
+}
+
+/** Whether proxies should be filtered by removing unconnected nodes regarding all on-screen nodes. */
+export class ProxyViewFilterUnconnectedToOnScreen implements RenderOption {
+    static readonly ID: string = "proxy-view-filter-unconnected";
+    static readonly NAME: string = "Filter Nodes Unconnected to On-Screen";
```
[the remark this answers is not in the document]
Eddykasp: We need to still do this on the server in semantics as well

packages/klighd-core/src/proxy-view/proxy-view-options.ts
```diff
@@ -0,0 +1,525 @@
...
+    static readonly DESCRIPTION: string = "Whether proxies should be interactable. Clicking on a proxy hops to its off-screen counterpart.";
+    static readonly DEFAULT: boolean = true;
+    readonly id: string = ProxyViewInteractiveProxies.ID;
+    readonly name: string = ProxyViewInteractiveProxies.NAME;
+    readonly type: TransformationOptionType = TransformationOptionType.CHECK;
+    readonly initialValue: boolean = ProxyViewInteractiveProxies.DEFAULT;
+    readonly description: string = ProxyViewInteractiveProxies.DESCRIPTION;
+    readonly renderCategory: string = ProxyViewCategory.ID;
+    currentValue = ProxyViewInteractiveProxies.DEFAULT;
+}
```
NiklasRentzCAU, line 164: I think this should be a debug option, the interactive features should always be turned on.

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,1530 @@
...
+        if (!point) {
+            // Point hasn't been stored yet, get parent position
+            point = this.getAbsolutePosition(node.parent as SKNode | SKEdge | SKPort | SKLabel);
+            point = Point.add(point, node.bounds);
+
+            // Also store this point
+            this.positions.set(id, point);
+        }
+        return point;
+    }
```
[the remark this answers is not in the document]
Eddykasp: the method doc already mentions this information

packages/klighd-core/src/actions/actions.ts
```diff
@@ -14,10 +14,11 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-import { ExportSvgAction, RequestExportSvgAction } from "sprotty";
+import { ExportSvgAction, RequestExportSvgAction, SGraph } from "sprotty";
```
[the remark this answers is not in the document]
Eddykasp: the organize imports command sorts it this way for some reason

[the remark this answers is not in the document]
Eddykasp: because it's correct, R does come before S in the alphabet

packages/klighd-core/src/proxy-view/filters/proxy-view-filter-options.ts
Eddykasp: similar to @command:klighd-vscode.data.clear in vscode, cli manually via cache

packages/klighd-core/src/proxy-view/proxy-view-util.ts
```diff
@@ -0,0 +1,1001 @@
...
+     * Note that `from` and `to` are not part of the returned path.
+     * @param from The bounds/point to route along the border from.
+     * @param fromBorder The border for `from`.
+     * @param to The bounds/point to route along the border to.
+     * @param toBorder The border for `to`.
+     * @param preferLeft Whether routing left should be preferred when `from` and `to` are vertically opposite of each other.
+     * @param preferTop Whether routing top should be preferred when `from` and `to` are horizontally opposite of each other.
+     * @returns A path along the border from `from` to `to`. Exclusive, e.g. (from, to).
+     */
+    export function routeAlongBorder(from: Bounds | Point, fromBorder: Bounds, to: Bounds | Point, toBorder: Bounds, preferLeft = true, preferTop = false): Point[] {
```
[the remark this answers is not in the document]
Eddykasp: non reachable code **is** reachable!

packages/klighd-core/src/proxy-view/proxy-view-util.ts
```diff
@@ -0,0 +1,1001 @@
...
+export function getIntersection(p1: Point, p2: Point, b: Bounds): Point | undefined {
+    // Intersection iff one of [p1, p2] in bounds and the other one out of bounds
+
+    if (Bounds.includes(b, p1)) {
+        // Intersection if p2 out of bounds
+        if (p2.x < b.x || p2.x > b.x + b.width) {
+            // Intersection at x, find y
+            const leftOrRight = p2.x < b.x ? b.x : b.x + b.width;
+
+            // Scalar of line equation, must be in [0,1] as to not be before p1 or after p2, could be ±inf
```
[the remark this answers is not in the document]
Eddykasp: @Eddykasp

packages/klighd-core/src/proxy-view/proxy-view-util.ts
```diff
@@ -0,0 +1,1001 @@
...
+
+/** Checks if `node` is selected or connected to any selected element. */
+export function isSelectedOrConnectedToSelected(node: SKNode): boolean {
+    const selectedNodes = SelectedElementsUtil.getSelectedNodes();
+    return node.selected || isConnectedToAny(node, selectedNodes);
+}
+
+//////// Classes ////////
+
+// TODO: GetSelectionAction
```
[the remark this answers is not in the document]
Eddykasp: @NiklasRentzCAU

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,1530 @@
...
+        styles.push({ color, type: K_BACKGROUND, selection: false });
+        styles.push({ color, type: K_FOREGROUND, selection: true });
+        styles.push({ color, type: K_BACKGROUND, selection: true });
+
+        if ("children" in clone) {
+            // Keep going recursively
+            (clone as any).children = this.changeColor((clone as any).children, ctx, color);
+        }
+
+        if ("junctionPointRendering" in clone) {
```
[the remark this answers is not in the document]
Eddykasp: @Eddykasp

packages/klighd-core/src/proxy-view/proxy-view.tsx
```diff
@@ -0,0 +1,1530 @@
...
+                // Change color
+                styles[i] = { ...styles[i], color };
+            }
+        }
+        styles.push({ color, type: K_FOREGROUND, selection: false });
+        styles.push({ color, type: K_BACKGROUND, selection: false });
+        styles.push({ color, type: K_FOREGROUND, selection: true });
+        styles.push({ color, type: K_BACKGROUND, selection: true });
+
+        if ("children" in clone) {
```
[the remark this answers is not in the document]
Eddykasp: @Eddykasp

packages/klighd-core/src/proxy-view/proxy-view.tsx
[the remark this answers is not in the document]
Eddykasp: `isPolyline` doesn't work as a replacement, proxies are broken then

Eddykasp:
To create a "signpost" projection of proxies onto edges (if they are on-screen), the way proxies are created and positioned has to be modified. In `proxy-view.ts` the method `getTransform` is responsible for positioning a node's proxy. The current structure only creates a single proxy for such any node. In a signpost projection we would require more. Additionally it needs to decided how to handle nodes where there currently is no connecting edge on screen. The simplest approach would be to not create a proxy in that case. But we could also use a radial projection for those nodes. Going further it might make sense for the synthesis to be able to specify using semantic tags, which projection should be used.

Eddykasp (approved):

PR #100, opened 2022-07-20
Reopen new diagram widget on start for current active editor.

soerendomroes:
Since the old webviewPanel does not refer to the old `SprottyWebview` (which is a `KlighdWebview`) cannot be readded to the list of available `KlighdWebview`s in `KlighdExtension` and the old panel cannot be used during the creation of a new `KlighdWebview` since it is `readonly`, we recreate the Diagram on the start of the extension for the currently active editor.

### DIFF
--- a/applications/klighd-vscode/src/extension.ts
+++ b/applications/klighd-vscode/src/extension.ts
@@ -3,7 +3,7 @@
  *
  * http://rtsys.informatik.uni-kiel.de/kieler
  *
- * Copyright 2021 by
+ * Copyright 2021-2022 by
  * + Kiel University
  *   + Department of Computer Science
  *     + Real-Time and Embedded Systems Group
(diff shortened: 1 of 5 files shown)


NiklasRentzCAU (commented):
It is very nice that it is possible to reopen the diagram on startup, albeit not using the VS Code API for that. For the SCCharts extension I can see this being the wanted behavior, as the diagrams always belong to the extension. As this is implemented here in klighd-vscode however, I would like this feature to have some slightly different behavior to avoid users that have this and an extension with very generic syntheses installed to always open (not reopen) a diagram on startup, even if they manually closed it in the previous session.

I would suggest to only restore the view when it was opened before (either via VS Code API or maybe via some persistent / session storage). Furthermore, what about this being an option to open a diagram automatically on startup or not?

applications/klighd-vscode/src/extension.ts
```diff
@@ -42,6 +42,7 @@ export function activate(context: vscode.ExtensionContext): void {
     // This extension should persist data in workspace state, so it is different for
     // each project a user opens. To change this, assign another Memento to this constant.
     const mementoForPersistence = context.workspaceState;
+    // const storage = new LocalStorage()
```
NiklasRentzCAU, line 45: Old debug statement?

NiklasRentzCAU (changes requested):
There is one debug statement left, please remove that. Otherwise, :+1:

NiklasRentzCAU (approved):
:+1:

PR #101, opened 2022-07-22
Pin sidebar action

soerendomroes:
Based on #100

### DIFF
--- a/packages/klighd-core/src/sidebar/sidebar-panel-registry.ts
+++ b/packages/klighd-core/src/sidebar/sidebar-panel-registry.ts
@@ -49,7 +49,7 @@ export class SidebarPanelRegistry extends Registry {
             if (this._currentPanelID === action.id && action.state === "show") return;
             if (this._currentPanelID !== action.id && action.state === "hide") return;
 
-            if (this._currentPanelID === action.id && !(this.currentPanel as SidebarPanel).panelPinned) {
+            if (this._currentPanelID === action.id) {
                 // Panel is active so it should either be hidden explicitly or toggled to be hidden
                 this._currentPanelID = null;
                 this.notifyListeners();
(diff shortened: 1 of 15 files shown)


packages/klighd-core/src/sidebar/sidebar-panel-registry.ts
```diff
@@ -49,7 +49,7 @@ export class SidebarPanelRegistry extends Registry {
             if (this._currentPanelID === action.id && action.state === "show") return;
             if (this._currentPanelID !== action.id && action.state === "hide") return;
 
-            if (this._currentPanelID === action.id) {
+            if (this._currentPanelID === action.id && !(this.currentPanel as SidebarPanel).panelPinned) {
```
NiklasRentzCAU, line 52: This seems wrong to dispatch a ToggleSidebarAction that does not toggle the sidebar. Does something speak against not dispatching the action in the first place if the panel is pinned in all cases where it could be toggled?

packages/klighd-core/src/sidebar/sidebar.tsx
```diff
@@ -141,4 +142,17 @@ export class Sidebar extends AbstractUIExtension {
...
+
+    /**
+     * Register a mouse left handler that hides the content if the mouse leaves the sidebar.
+     */
+    private addMouseLeaveListener(containerElement: HTMLElement): void {
+        containerElement.addEventListener("mouseleave", (e) => {
+            const currentPanelID = this.sidebarPanelRegistry.currentPanelID;
+            if (currentPanelID) {
+                this.actionDispatcher.dispatch(ToggleSidebarPanelAction.create(currentPanelID, "hide"));
+            }
```
NiklasRentzCAU, line 154: Actions should only be dispatched if they do something, this should have a guard checking for the panel being pinned.

NiklasRentzCAU (changes requested):
See the individual comments.

soerendomroes:
See #110

PR #103, opened 2022-07-25
Disable smart zoom per default.

soerendomroes:
We talked about this.


NiklasRentzCAU:
We talked about this again.
Superseded by #105.

PR #105, opened 2022-07-25
Improve smart zoom.

### LINKED ISSUE #104: Improve default smart zoom
Screenshot from 2022-07-25 10-24-30
Screenshot from 2022-07-25 10-24-58

Calculate opacity based on title color or overlay background.
(potentially also take the scalingfactor into account as an approximation to get better results in most cases)

Regions shouldn't be omitted in superstates as that is semantically strange.
In this case the synthesis should set a property that tells the client to not apply smart zoom to the superstates.

Placeholder symbol for hidden contents is now turned off by default, the rendering should be given by synthesis.

soerendomroes:
FIxes #104

### DIFF
--- a/packages/klighd-core/src/options/render-options-registry.ts
+++ b/packages/klighd-core/src/options/render-options-registry.ts
@@ -34,6 +34,7 @@ import { RangeOption, RenderOption, TransformationOptionType } from "./option-mo
     static readonly DEFAULT: boolean = true
     readonly id: string = ResizeToFit.ID;
     readonly name: string = ResizeToFit.NAME;
+    // readonly tooltip: string = ResizeToFit.TOOLTIP;
     readonly type: TransformationOptionType = TransformationOptionType.CHECK;
     readonly initialValue: boolean = ResizeToFit.DEFAULT;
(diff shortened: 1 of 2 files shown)


packages/klighd-core/src/options/render-options-registry.ts
```diff
@@ -34,6 +34,7 @@ import { RangeOption, RenderOption, TransformationOptionType } from "./option-mo
     static readonly DEFAULT: boolean = true
     readonly id: string = ResizeToFit.ID;
     readonly name: string = ResizeToFit.NAME;
+    // readonly tooltip: string = ResizeToFit.TOOLTIP;
```
NiklasRentzCAU, line 37: From another issue you were working on? Does not hurt, so I'll leave it in.

NiklasRentzCAU (approved):
Looks better than before :+1: 
For the zoom area placeholder, further details see #104.

Eddykasp (approved):
:+1:

Drakae (commented):
👍🏻

Drakae (approved):

PR #109, opened 2022-07-25
Semantic filtering client-side evaluation

Eddykasp:
https://github.com/kieler/KLighD/pull/137

### DIFF
--- a/packages/klighd-core/src/filtering/semantic-filtering-util.ts
+++ b/packages/klighd-core/src/filtering/semantic-filtering-util.ts
@@ -1,0 +1,318 @@
+/*
+ * KIELER - Kiel Integrated Environment for Layout Eclipse RichClient
+ *
+ * http://rtsys.informatik.uni-kiel.de/kieler
+ *
+ * Copyright 2022 by
+ * + Kiel University
+ *   + Department of Computer Science
+ *     + Real-Time and Embedded Systems Group
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+import { SKGraphElement } from "../skgraph-models"
+
+//// Base constructs ////
+
+/**
+ * Base interface for semantic filter rules.
+ */
+export interface SemanticFilterRule {
+    /** The rule name is used to identify rules and distinguish them from one another. */
+    ruleName?: string
+    /** The default value is used to indicate whether the semantic filter should be on or off by default. */
+    defaultValue?: boolean
+}
+
+/**
+ * A semantic filter tag is used as a filter rule that evaluates to true iff the tag is present
+ * on a graph element.
+ */
+export class SemanticFilterTag implements SemanticFilterRule {
+    ruleName?: string
+    tag: string
+    /** Not defining num defaults it to 0. */
+    num?: number
+}
+
+/**
+ * Base interface for connectives. Connectives take one or more filter rules as operands and 
+ * construct a new rule.
+ */
+export interface Connective extends SemanticFilterRule {
+    name: string
+}
+
+//// Connectives ////
+
+/**
+ * Base interface for unary connectives. Unary Connectives take exactly one operand.
+ */
+export interface UnaryConnective extends Connective {
+    operand: SemanticFilterRule
+}
+
+/**
+ * Base interface for binary connectives. Binary Connectives take exactly two operands.
+ */
+export interface BinaryConnective extends Connective {
+    leftOperand: SemanticFilterRule
+    rightOperand: SemanticFilterRule
+}
+
+/**
+ * Base interface for ternary connectives. Ternary Connectives take exactly three operands.
+ */
+export interface TernaryConnective extends Connective {
+    firstOperand: SemanticFilterRule
+    secondOperand: SemanticFilterRule
+    thirdOperand: SemanticFilterRule
+}
+
+//// Logic Connectives ////
+
+/**
+ * A Not Connective takes a rule R and evaluates to true
+ * iff
+ * R evaluates to false.
+ * @example !R
+ */
+export class NegationConnective implements UnaryConnective {
+    static NAME = "NOT"
+    name = NegationConnective.NAME
+    operand: SemanticFilterRule
+    ruleName?: string
+}
+
+/**
+ * An And Connective takes two rules R1 and R2 and evaluates to true
+ * iff
+ * R1 and R2 evaluate to true.
+ * @example R1 && R2
+ */
+export class AndConnective implements BinaryConnective {
+    static NAME = "AND"
+    name = AndConnective.NAME
+    leftOperand: SemanticFilterRule
+    rightOperand: SemanticFilterRule
+    ruleName?: string
+}
+
+/**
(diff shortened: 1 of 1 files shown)


packages/klighd-core/src/filtering/semantic-filtering-util.ts
```diff
@@ -0,0 +1,318 @@
...
+}
+
+/**
+ * A semantic filter tag is used as a filter rule that evaluates to true iff the tag is present
+ * on a graph element.
+ */
+export class SemanticFilterTag implements SemanticFilterRule {
+    ruleName?: string
+    tag: string
+    /** Not defining num defaults it to 0. */
```
NiklasRentzCAU, line 39: When creating such an element here would default to it being `undefined` and not `0`. I guess here you refer to tags from the server will default to `0`? Then I would remove the `?` here as a value is guaranteed.

packages/klighd-core/src/filtering/semantic-filtering-util.ts
```diff
@@ -0,0 +1,318 @@
...
+}
+
+//// Connectives ////
+
+/**
+ * Base interface for unary connectives. Unary Connectives take exactly one operand.
+ */
+export interface UnaryConnective extends Connective {
+    operand: SemanticFilterRule
+}
```
NiklasRentzCAU, line 58: A nullary connective for true/false would be good as well so their construction do not need to be hacky.

packages/klighd-core/src/filtering/semantic-filtering-util.ts
```diff
@@ -0,0 +1,318 @@
...
+ * iff
+ * R.num > correspondingTag.num.
+ * @example R.num > correspondingTag.num
+ */
+export class GreaterThanConnective implements UnaryConnective {
+    static NAME = "GREATERTHAN"
+    name = GreaterThanConnective.NAME
+    operand: SemanticFilterTag
+    ruleName?: string
+}
```
NiklasRentzCAU, line 192: Default relation definitions usually is using `<=`, `>=` and `=` resp. `!=` instead of `<` and `>`

packages/klighd-core/src/filtering/semantic-filtering-util.ts
```diff
@@ -0,0 +1,318 @@
...
+/**
+ * Creates a new filter with a function that can be applied to graph elements when given 
+ * a filter rule.
+ * @param rule the rule to construct the filter from
+ * @returns a new filter
+ */
+export function createFilter(rule: SemanticFilterRule): Filter {
+
+    let ruleName;
+    if (rule instanceof SemanticFilterTag) {
```
NiklasRentzCAU, line 228: Does this `instanceof` check work for rules defined by the server and serialized/deserialized? Usually the resulting objects are plain JS objects with fields and no such working `instanceof`. Usually this requires some [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

packages/klighd-core/src/filtering/semantic-filtering-util.ts
```diff
@@ -0,0 +1,318 @@
...
+        case IfThenConnective.NAME:
+            return !evaluateRule(binary.leftOperand, tags)
+                || evaluateRule(binary.rightOperand, tags);
+        case LogicEqualConnective.NAME:
+            return evaluateRule(binary.leftOperand, tags)
+                === evaluateRule(binary.rightOperand, tags);
+        case IfThenElseConnective.NAME:
+            return evaluateRule(ternary.firstOperand, tags)
+                ? evaluateRule(ternary.secondOperand, tags)
+                : evaluateRule(ternary.thirdOperand, tags);
```
NiklasRentzCAU, line 278: I think the evaluation of the directives would be better placed in the namespace of the directive classes in some `evaluate` function.

NiklasRentzCAU (commented):
As all of this (and the server changes) are very generic and do not do anything currently, some documentation how to use this, some tests, or simple examples should be added to show what this does.

packages/klighd-core/src/filtering/semantic-filtering-util.ts
Eddykasp: nullary connective is not allowed here due to linting rules as it would be empty

packages/klighd-core/src/filtering/semantic-filtering-util.ts
Eddykasp: this works

packages/klighd-core/src/filtering/semantic-filtering-util.ts
Eddykasp: could that not cause problems with the deserialization?

NiklasRentzCAU: serialization only uses fields and no methods of classes, I am not quite sure about the other way around. But [Actions](https://github.com/kieler/klighd-vscode/blob/main/packages/klighd-core/src/actions/actions.ts) circumvent this by putting the function into the namespace, not into some class.

Eddykasp: ah ok, I can do it that way

packages/klighd-core/src/filtering/semantic-filtering-util.ts
```diff
@@ -366,6 +373,74 @@ export namespace NumericNotEqualConnective {
...
+    name = NumericDividesConnective.NAME
+    leftOperand: SemanticFilterRule
+    rightOperand: SemanticFilterRule
+}
+
+export namespace NumericDividesConnective {
+    export function evaluate(conn: NumericDividesConnective): number {
+        return evaluateNumeric(conn.leftOperand) / evaluateNumeric(conn.rightOperand);
+    }
+}
```
NiklasRentzCAU, line 442: I suggest to rename these to …Addition…, …Subtraction…, …Multiplication…, and …Division…

packages/klighd-core/src/filtering/semantic-filtering-util.ts
```diff
@@ -0,0 +1,573 @@
...
+    export function evaluate(conn: NumericMultiplicationConnective): number {
+        return evaluateNumeric(conn.leftOperand) * evaluateNumeric(conn.rightOperand);
+    }
+}
+
+/**
+ * A Numeric Division Connective takes two numeric operands and evaluates
+ * to their product.
+ */
+ export class NumericDivisionConnective implements BinaryConnective {
```
fight4day, line 431: Should be quotient instead of product

packages/klighd-core/src/filtering/semantic-filtering-util.ts
```diff
@@ -0,0 +1,591 @@
...
+            return evaluateRule(rule, tags);
+        }
+    }
+
+}
+
+function evaluateNumeric(rule: SemanticFilterRule, tags: Array<SemanticFilterTag>): number {
+    // Rule is a Tag
+    if ((rule as SemanticFilterTag).tag !== undefined) {
+        let nodeTag = tags.find((tag: SemanticFilterTag) => tag.tag === (rule as SemanticFilterTag).tag)
```
fight4day, line 499: Should be const to pass checks

Eddykasp: :+1:

Eddykasp:
superseded by https://github.com/kieler/klighd-vscode/pull/86

PR #110, opened 2022-07-26
Render option documentation and quick action persistence

soerendomroes:


packages/klighd-core/src/options/option-models.ts
```diff
@@ -23,6 +23,12 @@ export interface RenderOption {
     initialValue: any;
     currentValue: any;
     description?: string;
+    /** The category this RenderOption is part of. */
+    renderCategory?: string;
+    /** The values this RenderOption has, if it's type is {@link TransformationOptionType.CHOICE}. */
+    renderChoiceValues?: any[];
```
NiklasRentzCAU, line 29: These are exactly the same as previously defined in the `SynthesisOption` interface a few lines below under the field names `category` and `values` (matching their Java counterparts). I suggest to rename these fields to `category` and `values` and remove the fields from the `SynthesisOption` interface for consistency between the option types and to remove duplicate fields on synthesis options.

packages/klighd-core/src/options/render-options-registry.ts
```diff
@@ -61,8 +80,35 @@ export class ShowConstraintOption implements RenderOption {
     readonly name: string = ShowConstraintOption.NAME;
     readonly type: TransformationOptionType = TransformationOptionType.CHECK;
     readonly initialValue: boolean = false;
+    readonly renderCategory: string = Appearance.ID
+    readonly description = 'Show marker for nodes that have layout constraints set.'
```
NiklasRentzCAU, line 84: …that have *interactive* layout constraints set

packages/klighd-core/src/options/render-options-registry.ts
```diff
@@ -71,11 +117,13 @@ export class ShowConstraintOption implements RenderOption {
  */
 export class UseSmartZoom implements RenderOption {
     static readonly ID: string = 'use-smart-zoom'
-    static readonly NAME: string = 'Smart Zoom'
+    static readonly NAME: string = 'Enable Smart Zoom'
```
NiklasRentzCAU, line 120: I think in the semantics it was discussed at some point that no options should have "show", "enable" or similar in their option label, as having the checkbox should clarify that this option is enabled when it is checked. Furthermore, the description should now further clarify that if the user does not immediately understand what the option does from its name.

packages/klighd-core/src/options/render-options-registry.ts
```diff
@@ -132,6 +186,8 @@ export class SimplifySmallText implements RenderOption {
     readonly name: string = SimplifySmallText.NAME
     readonly type: TransformationOptionType = TransformationOptionType.CHECK
     readonly initialValue: boolean = true
+    readonly renderCategory: string = SmartZoom.ID
+    readonly description = 'Whether small text is simplified to colored rectangles.'
```
NiklasRentzCAU, line 190: ~small~ illegible

packages/klighd-core/src/options/render-options-registry.ts
```diff
@@ -152,6 +208,8 @@ export class TextSimplificationThreshold implements RangeOption {
     }
     readonly stepSize = 0.1
     readonly initialValue: number = TextSimplificationThreshold.DEFAULT
+    readonly renderCategory: string = SmartZoom.ID
+    readonly description = 'The threshold font size to simplify text.\nIf set to 3 a text which is 3 or less pixel high is simplified.'
```
NiklasRentzCAU, line 212: For future reference: I think this option should be a debug option (possibility for that implemented in tik's bt code) with a reasonable value for illegible text sizes.

packages/klighd-core/src/options/render-options-registry.ts
```diff
@@ -173,6 +231,8 @@ export class TitleScalingFactor implements RangeOption {
     }
     readonly stepSize = 0.01
     readonly initialValue: number = TitleScalingFactor.DEFAULT
+    readonly renderCategory: string = SmartZoom.ID
+    readonly description = 'Factor to scale region titles compared to their original size.\nIf set to 1 a region title will never exceed their original size.'
```
NiklasRentzCAU, line 235: If set to 1 a region title will never be smaller than its original apparent size. This wording sounds as if the text will always be smaller than original.

packages/klighd-core/src/options/render-options-registry.ts
```diff
@@ -206,22 +268,37 @@ export class MinimumLineWidth implements RangeOption {
...
+    static readonly DEFAULT: ShadowOption = ShadowOption.KIELER_STYLE
+    readonly id: string = Shadows.ID
+    readonly name: string = Shadows.NAME
+    readonly type: TransformationOptionType = TransformationOptionType.CHOICE
+    readonly initialValue: ShadowOption = Shadows.DEFAULT
+    readonly renderCategory: string = Appearance.ID
+    readonly renderChoiceValues? = [ShadowOption.PAPER_MODE, ShadowOption.KIELER_STYLE]
+    readonly description = 'The style shadows should be drawn in, either the paper mode shadows (nice, but slow in performance)'
+     + 'or in default KIELER Style (fast, not as nice looking).'
+     + 'KIELER Style multiple shapes in form of the node behind it.'
```
NiklasRentzCAU, line 299: missing verb

packages/klighd-core/src/sidebar/sidebar-panel-registry.ts
```diff
@@ -43,6 +46,22 @@ export class SidebarPanelRegistry extends Registry {
...
+        // Reopen general panel if a panel was pinned.
+        // Record has to be retrieved manually since the renderOptionsRegistry is not yet initialized and
+        // cannot be changed to distribute a ready signal.
+        this.storage.getItem<Record<string, unknown>>('render').then((data: Record<string, unknown>) => {
+            for (const entry of Object.entries(data)) {
+                if (entry[0] == PinSidebarOption.ID) {
+                    if (entry[1] && this.allPanels.length > 0) {
+                        this._currentPanelID = this.allPanels[0].id
+                    }
+                }
```
NiklasRentzCAU, line 60: Why aren't these nested if's merged into one?

NiklasRentzCAU (changes requested):
See the comments. This will be a very nice QoL change :)

packages/klighd-core/src/sidebar/sidebar-panel-registry.ts
soerendomroes: Ok

packages/klighd-core/src/options/render-options-registry.ts
soerendomroes: Ok

packages/klighd-core/src/options/render-options-registry.ts
soerendomroes: But if the space does not permit it, the region title is drawn smaller. As I see it we specify an upper bound here. I will change it to will be their original size if the space permits it.

packages/klighd-core/src/options/render-options-registry.ts
soerendomroes: I suggest making a ticket for this. Here is a bad spot for future references. #112

packages/klighd-core/src/options/render-options-registry.ts
soerendomroes: ok

packages/klighd-core/src/options/render-options-registry.ts
soerendomroes: Ok

packages/klighd-core/src/options/render-options-registry.ts
soerendomroes: ok

packages/klighd-core/src/options/option-models.ts
soerendomroes: But the fields are not the same a renderCategory is only a string since the render option registration process does not allow to add a singleton category option on the client side.

soerendomroes: Maybe RenderOption is just not a good super type. I suggest to make a general option type and extend it by renderCategory and category for RenderOptions and SynthesisOptions

packages/klighd-core/src/options/general-panel.tsx
```diff
@@ -121,7 +141,13 @@ export class GeneralPanel extends SidebarPanel {
...
                                 title={action[1]}
                                 class-options__icon-button="true"
-                                on-click={() => this.handleQuickActionClick(action[0])}
+                                class-sidebar__enabled-button={!!action[4]}
+                                on-click={() => {
+                                    if (action[5]) {
+                                        action[5].apply(this)
+                                    }
+                                    this.handleQuickActionClick(action[0])
+                                }}
```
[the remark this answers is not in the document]
soerendomroes: Ok

packages/klighd-core/src/options/option-models.ts
NiklasRentzCAU: I see, the category serialized from the server side is also not serialized into a singleton, currently causing multiple equal categories in the option data, of which only the string ID will be used anyway. I guess leaving it like this is okay, as 'correctly' using an option as the category with just `{id: theId} as SynthesisOption` could potentially lead to other problems.

But the `choiceValues`/`values` fields should be and do the same thing I think

soerendomroes: Yes, I did it that way.

NiklasRentzCAU (approved):
:+1:

PR #111, opened 2022-07-27
Open diagram if an editor is eventually selected.

soerendomroes:
The diagram did not open if initially, no active text editor existed. Now the diagram opens when a text editor is selected and not if one was selected on startup.
(Everything or course only if the diagram was not closed)

### DIFF
--- a/applications/klighd-vscode/src/klighd-webview-reopener.ts
+++ b/applications/klighd-vscode/src/klighd-webview-reopener.ts
@@ -15,14 +15,14 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 
-import { commands, Uri, window } from "vscode";
+import { commands, Disposable, Uri, window } from "vscode";
 import { command } from "./constants";
 import { StorageService } from "./storage/storage-service";
 
 export class KlighdWebviewReopener {
 
     private readonly storage: StorageService
-    private toDispose: { dispose(): any }[] = []
+    private toDispose: Disposable[] = []
 
     constructor(storage: StorageService) {
         this.storage = storage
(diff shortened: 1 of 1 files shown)


applications/klighd-vscode/src/klighd-webview-reopener.ts
```diff
@@ -22,20 +22,31 @@ import { StorageService } from "./storage/storage-service";
 export class KlighdWebviewReopener {
 
     private readonly storage: StorageService
+    private toDispose: { dispose(): any }[] = []
```
NiklasRentzCAU, line 25: Interesting way to unregister a function from an event, had to look it up in the VS Code API to understand what is going on here. There I found that the event that the function is added to returns (in fact, any `Event`) returns a `Disposable`, so couldn't this `toDispose` field be just a `Disposable[]`?

applications/klighd-vscode/src/klighd-webview-reopener.ts
```diff
@@ -22,20 +22,31 @@ import { StorageService } from "./storage/storage-service";
...
+                const uri = activeTextEditor.document.fileName
                 commands.executeCommand(command.diagramOpen, Uri.file(uri))
+            } else {
+                // Register this an active editor changed to open the diagram then.
+                this.toDispose.push(window.onDidChangeActiveTextEditor(editor => {
+                    let uri = undefined
+                    if (editor) {
+                        uri = editor.document.uri
+                    }
+                    commands.executeCommand(command.diagramOpen, uri)
```
NiklasRentzCAU, line 45: So you call the diagramOpen command with `undefined` if the active editor changes to `undefined`? What does that even do?

> *Note* that the event also fires when the active editor changes to `undefined`.

soerendomroes: I don't care and this will not happen since it cannot be called if the editor is undefined.

applications/klighd-vscode/src/klighd-webview-reopener.ts
soerendomroes: Ok

NiklasRentzCAU (approved):
:+1:

PR #116, opened 2022-08-22
Values are now shown while the slider is held

### LINKED ISSUE #108: Sliders in sidebar should show numbers during slide movement and not only after the mouse is let go
image

fight4day:
Resolves #108


Eddykasp (changes requested):
This causes a new request to be sent to the server for each change, which we don't want. The request should only be sent once the slider is let go.

Eddykasp (approved):
Looks good to me! Any further concerns @NiklasRentzCAU ?

NiklasRentzCAU (approved):
:+1:

PR #130, opened 2023-03-15
child placement and rendering bounding box fixes

### LINKED ISSUE #68: Incorrect handling of padding in rendering
Setting the padding of a parent node to 0 should result in the child node being placed in the same origin as its parent. This appears to be a rendering issue in the VS Code extension. Below is an example diagram as well as the expected result (as it is in Kieler) and the actual result in VS Code.

The position of the child node is wrong and the actual size is also too large.

## Sample SCChart
```
scchart PaddingExample {
	
    @layout[elk.padding] "[top=0.0,left=0.0,bottom=0.0,right=0.0]"
    region reg "" {
        initial state SystemStateWithLongName
    }	
}
```

## Result in Kieler
padding_kieler
## Result in the VSCode Extension
padding_vscode

### LINKED ISSUE #82: Calculate absolute coordinates on server and store directly in sknodes
Several features, such as incremental diagram generation, to an extent the depth map and the future proxy-view require the absolute coordinates of nodes and compute them independently. This redundancy should be extracted and centralized, possibly to the language server.

NiklasRentzCAU:
Fixes the child offsets for graphs with offset child areas.
Fixes the width and height of ellipses and rectangles to represent the correct bounding box of the renderings, not the middle points of their strokes.

Fixes #68
May fix #82

### DIFF
--- a/packages/klighd-core/src/views-common.ts
+++ b/packages/klighd-core/src/views-common.ts
@@ -571,6 +571,42 @@ export function transformationToSVGString(transformation: Transformation): strin
 }
 
 /**
+ * Reverses an array of transformations such that applying the transformation and its reverse counterpart will result in the identity transformation.
+ * @param transformations The transformations to reverse.
+ * @returns The reversed transformations.
+ */
+export function reverseTransformations(transformations: Transformation[]): Transformation[] {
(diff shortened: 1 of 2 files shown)


Eddykasp (approved):
:+1:

PR #145, opened 2023-06-07
Core: switched tabs + added quickbar for options

TheRichmend:
1. switched tabs, so that options are at first, general second, and bookmarks at last.
2. also gave them static values, because general was at -10 before and the others at default (0)
3. let sidebar-panel init the quickactionbar. now options and general inherit it. 
--> now options has a quickactionbar aswell.

### DIFF
--- a/packages/klighd-core/src/bookmarks/bookmark-panel.tsx
+++ b/packages/klighd-core/src/bookmarks/bookmark-panel.tsx
@@ -31,8 +31,8 @@ import { Bookmark, GoToBookmarkAction, CreateBookmarkAction, DeleteBookmarkActio
 @injectable()
 export class BookmarkPanel extends SidebarPanel {
     // sets the bookmarkpanel at the bottom
-    // hirarchy is: first = -10; middle = 0; last = 10;
-    readonly position = 10; // --> last position
+    // hierarchy is: first elem has the lowest number. so the last one got the highest
+    readonly position = 10; // --> last position (at the moment)
     
     @inject(DISymbol.BookmarkRegistry) private bookmarkRegistry: BookmarkRegistry
 
--- a/packages/klighd-core/src/bookmarks/bookmark-panel.tsx
+++ b/packages/klighd-core/src/bookmarks/bookmark-panel.tsx
@@ -4,7 +4,7 @@
  *
  * http://rtsys.informatik.uni-kiel.de/kieler
  * 
- * Copyright 2021 by
+ * Copyright 2021-2023 by
  * + Kiel University
  *   + Department of Computer Science
  *     + Real-Time and Embedded Systems Group
@@ -18,7 +18,7 @@
 
 import { inject, injectable, postConstruct } from "inversify";
 import { VNode } from "snabbdom";
-import { html, IActionDispatcher, TYPES } from "sprotty"; // eslint-disable-line @typescript-eslint/no-unused-vars
+import { html } from "sprotty"; // eslint-disable-line @typescript-eslint/no-unused-vars
 import { DISymbol } from "../di.symbols";
 import { FeatherIcon } from '../feather-icons-snabbdom/feather-icons-snabbdom';
 import { SidebarPanel } from "../sidebar";
@@ -30,8 +30,10 @@ import { Bookmark, GoToBookmarkAction, CreateBookmarkAction, DeleteBookmarkActio
  */
 @injectable()
 export class BookmarkPanel extends SidebarPanel {
-
-    @inject(TYPES.IActionDispatcher) private actionDispatcher: IActionDispatcher;
+    // sets the bookmarkpanel at the bottom
+    // hirarchy is: first = -10; middle = 0; last = 10;
+    readonly position = 10; // --> last position
+    
     @inject(DISymbol.BookmarkRegistry) private bookmarkRegistry: BookmarkRegistry
 
     @postConstruct()
--- a/packages/klighd-core/src/options/general-panel.tsx
+++ b/packages/klighd-core/src/options/general-panel.tsx
@@ -51,7 +51,7 @@ export class GeneralPanel extends SidebarPanel {
         this.preferencesRegistry.onChange(() => this.update());
         this.renderOptionsRegistry.onChange(() => this.update());
 
-        this.assignQuickActions();
+        this.assignQuickActions()
     }
 
     get id(): string {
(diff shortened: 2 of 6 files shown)


packages/klighd-core/src/bookmarks/bookmark-panel.tsx
```diff
@@ -30,8 +30,10 @@ import { Bookmark, GoToBookmarkAction, CreateBookmarkAction, DeleteBookmarkActio
  */
 @injectable()
 export class BookmarkPanel extends SidebarPanel {
-
-    @inject(TYPES.IActionDispatcher) private actionDispatcher: IActionDispatcher;
+    // sets the bookmarkpanel at the bottom
+    // hirarchy is: first = -10; middle = 0; last = 10;
```
NiklasRentzCAU, line 34: typo: "hierarchy"

packages/klighd-core/src/bookmarks/bookmark-panel.tsx
```diff
@@ -30,8 +30,10 @@ import { Bookmark, GoToBookmarkAction, CreateBookmarkAction, DeleteBookmarkActio
...
+    readonly position = 10; // --> last position
```
NiklasRentzCAU, line 35: this would not be last, if another panel was added with a bigger number, comment could be more generic here (higher number leads to lower priority or similar)

packages/klighd-core/src/options/general-panel.tsx
```diff
@@ -16,42 +16,32 @@
...
-    // This panel should always have the first trigger in the sidebar
-    readonly position = -10;
-
-    /** Quick actions reference for this panel */
-    private quickActions: QuickActionOption[];
-
-    @inject(TYPES.IActionDispatcher) private actionDispatcher: IActionDispatcher;
+    // Sets this panel at the second position
+    // hirarchy is: first = -10; middle = 0; last = 10;
+    readonly position = 0; // --> middle position
```
NiklasRentzCAU, line 41: see above

packages/klighd-core/src/options/general-panel.tsx
```diff
@@ -72,62 +62,9 @@ export class GeneralPanel extends SidebarPanel {
...
-                state: this.renderOptionsRegistry.getValue(PinSidebarOption)
-            },
-        ];
-    }
-
     update(): void {
-        this.assignQuickActions()
-        super.update()
+        super.assignQuickActions();
+        super.update();
```
NiklasRentzCAU, line 67: Usually we have coding guidelines enforced by a Plugin called "Prettier", which somehow is not enabled in this repository yet - I may have a look at that later. One of the rules is that we do not allow the usage of these optional semicolons, so you should not explicitly add them where there are none before.

packages/klighd-core/src/options/general-panel.tsx
```diff
@@ -136,7 +73,7 @@ export class GeneralPanel extends SidebarPanel {
                 <div class-options__section="true">
                     <h5 class-options__heading="true">Quick Actions</h5>
                     <div class-options__button-group="true">
-                        {this.quickActions.map((action) => (
+                        {this.getQuickAction().map((action) => (
```
NiklasRentzCAU, line 76: should be called as the plural `getQuickActions()`

packages/klighd-core/src/options/options-panel.tsx
```diff
@@ -28,12 +28,20 @@ import { SidebarPanel } from "../sidebar";
 /** Sidebar panel that displays server provided KLighD options.  */
 @injectable()
 export class OptionsPanel extends SidebarPanel {
+    // sets this panel at the top position
+    // hirarchy is: first = -10; middle = 0; last = 10;
+    readonly position = -10;  // --> first position                                                            
     @inject(DISymbol.OptionsRegistry) private optionsRegistry: OptionsRegistry;
     @inject(DISymbol.OptionsRenderer) private optionsRenderer: OptionsRenderer;
-
+
```
NiklasRentzCAU, line 36: unnecessary whitespace change, try to avoid these

packages/klighd-core/src/options/options-panel.tsx
```diff
@@ -28,12 +28,20 @@ import { SidebarPanel } from "../sidebar";
...
     @inject(DISymbol.OptionsRegistry) private optionsRegistry: OptionsRegistry;
     @inject(DISymbol.OptionsRenderer) private optionsRenderer: OptionsRenderer;
-
+    
     @postConstruct()
     init(): void {
         this.optionsRegistry.onChange(() => this.update());
+        this.assignQuickActions();
+    }
+    update(): void {
```
NiklasRentzCAU, line 42: always have an empty line between methods (should usually also be enforced by prettier)

packages/klighd-core/src/sidebar/sidebar-panel.ts
```diff
@@ -15,9 +15,21 @@
...
+//rik imports
+import { RefreshDiagramAction } from "@kieler/klighd-interactive/lib/actions";
+import { CenterAction } from "sprotty-protocol";
+import { KlighdFitToScreenAction, RefreshLayoutAction } from "../actions/actions";
+import { CreateBookmarkAction } from "../bookmarks/bookmark";
+import { PossibleQuickAction, QuickActionOption } from "../options/option-models";
+import { SetRenderOptionAction } from "../options/actions";
+import { PinSidebarOption,RenderOptionsRegistry, ResizeToFit } from "../options/render-options-registry";
+import { IActionDispatcher, RequestExportSvgAction, TYPES } from "sprotty";
+import { DISymbol } from "../di.symbols";
```
NiklasRentzCAU, line 30: no need to mention that these imports were added by you.
Please also sort the import lines alphabetically, i.e. sort "sprotty" above to "sprotty-protocol". See other files for reference.

packages/klighd-core/src/sidebar/sidebar-panel.ts
```diff
@@ -84,4 +101,72 @@ export abstract class SidebarPanel implements ISidebarPanel {
...
+    public getQuickAction() :QuickActionOption[]{
+        return this.quickActions;
+    }
+    protected handleQuickActionClick(type: PossibleQuickAction): void {
+        const action = this.getQuickAction().find((a) => a.key === type)?.action;
+
+        if (!action) return;
+
+        this.actionDispatcher.dispatch(action);
+    }
```
NiklasRentzCAU, line 168: Methods should be protected here imo.
Methods need documentation like this:
```
/**
 * explain what this method does and what it returns. See other documentation
 */
```

please remember to add one empty line between methods.

packages/klighd-core/src/sidebar/sidebar-panel.ts
```diff
@@ -84,4 +101,72 @@ export abstract class SidebarPanel implements ISidebarPanel {
...
+
+
```
NiklasRentzCAU, line 170: But no double newlines ;)

NiklasRentzCAU (changes requested):
Looks good! Please do the requested changes and re-request the PR once your are done.

NiklasRentzCAU (changes requested):
I just improved the last remaining style and texts, you may have a look at what you missed in my commit here.

While testing this I noticed the 'Pin Sidebar' and 'Fit to screen' actions (so the ones with a persisting state) do not work as before anymore when changing between the tabs: updating it in one tab causes the other to not notice the change, causing it to behave in an unwanted manner or get stuck altogether. Please review the functionality of the buttons so that the state changes are persisted between the sidebar tabs.

NiklasRentzCAU (changes requested):
A few last comments, then I am (hopefully) happy with this PR.
Please note to resolve the conversations of my reviews after you fixed them. That way I can better see you worked on these comments directly.

Finally I will squash the commits of this PR into a single clean commit into the main branch, so no worries of some back-and-fourths in the code, the final commit on the main branch will look clean.

NiklasRentzCAU (approved):
I just changed a few last things I meant with the last review, now really making the quick actions bar a reusable jsx component. Now, this looks good, thank you for your help!

PR #146, opened 2023-07-20
Core: Sidebar DropDownMenu

TheRichmend:
I changed single choice options to a dropdown menu. therefore i changed the file of option-inputs from a .ts into .tsx.

### DIFF
--- a/packages/klighd-core/src/options/components/option-inputs.tsx
+++ b/packages/klighd-core/src/options/components/option-inputs.tsx
@@ -56,27 +56,27 @@ interface ChoiceOptionProps extends BaseProps<string> {
     availableValuesLabels?: string[];
 }
 
+/** This Function calls the onChange method, so that a click cause an active change */
+function getCurrentValue(a: any, props: ChoiceOptionProps) {
+   const value = a.target.value
+   props.onChange(value)
+  }
+
 /** Render a labeled group of radio inputs. */
 export function ChoiceOption(props: ChoiceOptionProps): VNode {
     // The sprotty jsx function always puts an additional 'props' key around the element, requiring this hack.
     props = (props as any as {props: ChoiceOptionProps}).props
-    return (
+    return ( 
         <div class-options__input-container="true">
-            <legend>{props.name}</legend>
-            {props.availableValues.map((value, i) => (
-                <label key={value} htmlFor={props.availableValuesLabels?.[i] ?? value} title={props.description ?? props.name}>
-                    <input
-                        class-options__input="true"
-                        type="radio"
-                        title={props.description ?? props.name}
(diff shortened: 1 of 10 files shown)


packages/klighd-core/src/options/components/option-inputs.tsx
```diff
@@ -56,26 +56,32 @@ interface ChoiceOptionProps extends BaseProps<string> {
     availableValuesLabels?: string[];
 }
 
+/** This Function calls the onChange method, so that a click cause an active change */
+function getCurrentValue(e: Event, props: ChoiceOptionProps) {
```
NiklasRentzCAU, line 60: Both the comment nor the function name really reflect what this function does. Always try to think of what the method calculates or executes when looking at it alone, without necessarily knowing about the context it is called from in your implementation. The apparent use of this method is to call the `onChange` function of the given `props` for the given event, if applicable. From this method alone this has nothing to do with "clicks" directly. and `getCurrentValue` sounds like a getter of some sorts, a better name would be `handleChoiceEvent`, `onChoiceEvent` or similar.
Furthermore, we like to include the return type on every method, so this should state `: void` before the opening bracket `{`.

packages/klighd-core/src/options/components/option-inputs.tsx
```diff
@@ -56,26 +56,32 @@ interface ChoiceOptionProps extends BaseProps<string> {
...
+    if (e.target == null) {
+        return
+    }
+    if (!(e.target instanceof HTMLSelectElement)) {
+        return
+    }
+    const val = e.target.value
+    return props.onChange(val)
```
NiklasRentzCAU, line 68: such small otherwise unused expressions can be included in the method call directly, i.e. `props.onChange(e.target.value)`.
remove the return here, a void method does not need to conclude with a return in the end and should also not return some value.

NiklasRentzCAU (changes requested):
Please re-open a new PR with the changes of this one on a new branch. You based this PR off of your implemented changes of the previous PR and not the main branch, thus causing the changes of the other PR to be included in this PR as well. Usually Git can manage that, but as I squashed the PR into the main, thus not preserving the 9 commits of your previous PR, git sees these as new Commits. Merging this here as well would leave the history in a weird state with duplicated changes on the same files. Committing only the new changes on a new branch or cherry-picking them from this branch on a new branch from the current main will clean up that history. Please do that before merging the changes.

Eddykasp:
I believe this PR is superseded by https://github.com/kieler/klighd-vscode/pull/151 and can be closed, correct?

NiklasRentzCAU:
Superseded by #151.

PR #150, opened 2023-08-07
Update to Sprotty 0.12

NiklasRentzCAU:
As the update to Sprotty 0.13 did cause some problems, at least update to 0.12 for now. This includes all necessary changes for that update branched from the first changes of the 0.13 branch.


soerendomroes (commented):
If you have tested it, I would approve the changes.

soerendomroes (approved):

PR #151, opened 2023-09-07
Rik/core: dropdownmenu

TheRichmend:
this is an implementation of a dropdownmenu inside the sidebar for a couple of options

### DIFF
--- a/packages/klighd-core/styles/options.css
+++ b/packages/klighd-core/styles/options.css
@@ -168,7 +168,7 @@
 
 .options__selection {
     padding: 4px 6px;
-
+    width: 200px;
     border: 2px solid var(--kdc-color-sidebar-trigger-background-hover);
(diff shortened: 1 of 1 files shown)


NiklasRentzCAU:
You did not include [this](https://github.com/kieler/klighd-vscode/blame/42e250fa00c9999da21d7a1293319a38816f1500/packages/klighd-core/styles/options.css#L171) change in this PR, was that necessary / what did that change? Otherwise, ok.

TheRichmend:
that was the parameter, that restricted the width of the dropdown-line. i have added it now

NiklasRentzCAU (approved):
:+1:

NiklasRentzCAU (changes requested):
As discussed in the last meeting, we want this behavior as an option to choose by the synthesis instead of it being a new default. See #155 for details.

PR #153, opened 2023-09-08
support for top-down scalefactor

Eddykasp:
corresponding server side PR https://github.com/kieler/KLighD/pull/173 but doesn't rely on it to not break things


packages/klighd-core/src/depth-map.ts
```diff
@@ -172,7 +172,18 @@ export class DepthMap {
...
         if (element.parent === element.root && element instanceof KNode) {
             const providedRegion = new Region(element)
-            providedRegion.absolutePosition = element.bounds
+
+            let topdownScaleFactor = 1
+            if ((element.parent as any).properties == undefined || (element.parent as any).properties['org.eclipse.elk.topdown.scaleFactor'] == undefined) {
+                topdownScaleFactor = 1
+            } else {
+                topdownScaleFactor = (element.parent as any).properties['org.eclipse.elk.topdown.scaleFactor']
+            }
```
NiklasRentzCAU, line 181: the scale factor already is 1, so just inverting the condition would make an else clause unnecessary. Or, why not write this as a one-liner with a fallback like this: `const topdownScaleFactor = (element.parent as any).?properties['org.eclipse.elk.topdown.scaleFactor'] ?? 1`

And: is this `any` cast necessary?

(multiple occurances of this in this PR)

packages/klighd-core/src/depth-map.ts
```diff
@@ -369,6 +386,7 @@ export class DepthMap {
      */
     computeDetailLevel(region: Region, viewport: Viewport, relativeThreshold: number, scaleThreshold: number): DetailLevel {
         if (!this.isInBounds(region, viewport)) {
+            console.log(region.boundingRectangle.id + " out of bounds")
```
NiklasRentzCAU, line 389: debug log or message important for the user?

packages/klighd-core/src/depth-map.ts
Eddykasp: good suggestion for simplifying the code,  I'll have to check the any cast

packages/klighd-core/src/depth-map.ts
Eddykasp: I think this was for debugging and will be removed

PR #156, opened 2023-09-29
allow tests with jsx dependencies

NiklasRentzCAU:
This changes some compiler options, especially to allow mocha to test .tsx files or test files transitively requiring other .tsx files.

This PR does not break anything in the current running system, but we should consider upgrading more versions and maybe the typescript compiler settings in the whole project, as started in the testS branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
         "watch": "lerna run watch --parallel",
         "package": "lerna run build && lerna run package --parallel",
         "distribute": "lerna run build && lerna run distribute --parallel",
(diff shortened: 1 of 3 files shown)


soerendomroes (approved):
If it works...
Why the change to es2017?

NiklasRentzCAU:
Executing a test on something in `views-common` without that change states that something is using a feature that is only available from es2017, so I upgraded to that. That is why I would want to upgrade everything some time in the future.

PR #158, opened 2023-10-17
Added sidebar icon size css class as a workaround.

soerendomroes:


NiklasRentzCAU (approved):
If this fixes the problem in Firefox, fine by me.

PR #181, opened 2024-07-19
Add smart zoom support for top-down layout

Eddykasp:
Also improves the position calculation in smart zoom.


soerendomroes (approved):
This looks like a very good solution.
Your parents and coworkers are probably proud of your exceptional skill.

PR #187, opened 2024-07-25
Allow KImages to refer to arbitrary URIs, which will be linked in the SVG

NiklasRentzCAU:
A first implementation to include images in the SVG outside of the bundled language server jar in the diagram. URIs available to the application can be used with this, e.g. URLs to images found online.
To refer to a URI image, add a KImage to the model with `bundleName = "URI"` and the URI in the `imagePath` attribute.

This implementation just refers to those images, a different implementation also enabling to refer to e.g. image files next to the visualized model will be worth expanding on this feature, as well as downloading and embedding the images in the SVG instead of linking to them for better compatibility during SVG export.

### DIFF
--- a/packages/klighd-core/src/diagram-server.ts
+++ b/packages/klighd-core/src/diagram-server.ts
@@ -3,7 +3,7 @@
  *
  * http://rtsys.informatik.uni-kiel.de/kieler
  *
- * Copyright 2021 by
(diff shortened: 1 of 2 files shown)


NiklasRentzCAU:
This is a first implementation towards the ticket kieler/klighd#146.

Issue #139, opened 2023-04-25
Client-only Micro/Macro Layout / Synthesis


NiklasRentzCAU:
For the future, we want to allow to use KLighD technology while not needing to have a Java-based Server to improve performance and widen the use cases for KLighD. In a first step this includes using Langium as a DSL framework and use KLighD for such DSLs as well.

Current steps necessary for implementing the client-only workflow:
See [here](https://rtsys.informatik.uni-kiel.de/confluence/x/DAAeC) for some internal details.
![photo_2022-08-25_10-55-44](https://user-images.githubusercontent.com/47597619/234217248-f66fcba8-5f8a-444a-8e41-7761fa3b2cd6.jpg)

- [ ] elkjs usable with KLighD-JS
- [ ] (together with step 1) micro layout on the client
- [ ] simple (Xtext/**Langium**) model to SKGraph synthesis on the client
And to make writing more complex syntheses easier:
- [ ] helper code to write syntheses on the client

NiklasRentzCAU:
Macro layout on the client is now possible. An issue related to that yet persists: 
- macro layout only supports the layered algorithm
calling elkjs with any other algorithm causes it to fail layout with a message like this one:
`Error: org.eclipse.elk.core.UnsupportedConfigurationException: Layout algorithm '"radial"' not found for Root Node file:///home/path-to-file/small.kgt > Node $root`

NiklasRentzCAU:
A proof of concept and a starting point in the code together with the necessary steps for implementation have been added to the branch [nre/client-micro-layout](https://github.com/kieler/klighd-vscode/tree/nre/client-micro-layout) in this repository.

PR #195, opened 2024-09-28
implement client-only layout.

NiklasRentzCAU:
Hardcode to use it for now, switch between modes missing yet.


Eddykasp (commented):
We should import all of elks algorithms not just layered

NiklasRentzCAU:
Before merging, this PR needs to include a way to switch between server and client layout, with the default being server layout for now.
For further improvement, this PR could also include a new message part in the "capabilities" messaging that asks the server if it supports server layout (to which the Java server would reply true) and automatically adapt its setting to the supported one (client layout if the server does not support it, server layout otherwise)

Eddykasp:
- [ ] add json schema for our lsp messages
- [ ] add validification using the new json schemas

NiklasRentzCAU:
Closes #139

Eddykasp (commented):
I think this is still missing the change to make the debug option switch no longer a debug option

packages/klighd-core/src/options/render-options-registry.ts
```diff
@@ -509,8 +509,6 @@ export class DebugOptions implements RenderOption {
     readonly description = 'Whether debug options should be shown.'
 
     currentValue = false
-
-    debug = true
```
NiklasRentzCAU, line 513: "DebugOptions" is no longer a debug option, see here.

Eddykasp (approved):
lgtm

PR #162, opened 2024-01-09
Version Updates

NiklasRentzCAU:
Updates Sprotty to version 1.1.0 and alongside it now requires node 16, es2017, new webpack 5 and others.

PR #200, opened 2024-10-22
Removes the double diagram request on opening VS Code with no editor in focus

NiklasRentzCAU:
During adoption to Sprotty 0.13 another way of updating the diagram was introduced that always requests a diagram when the editor is switched. When opening VS Code now with no editor in focus (e.g. when the diagram was in focus, which is not re-opened by VS Code by default), clicking on an editor triggered this newly introduced switch AND this now-removed code we had for exactly that use case. As this is now superfluous and causes duplicated requests, this removes that issue.
See #162


Eddykasp (approved):
thanks for fixing this!

PR #206, opened 2024-11-15
Create build-types.yml

### LINKED ISSUE #205: Add Github action to test whether schema to type generation works
To ensure that we don't commit broken schemas to the master we should add an action to automaticallly generate types for languages that we are interested. E.g. for nor pydantic types using datamodel-codegen as required for [plyghd](https://github.com/kieler/plyghd-ls-demonstrator)

Eddykasp:
This proposed action should attempt to generate pydantic types from the defined json schemas to catch any breaking changes before they are pushed to the master.

closes #205

### DIFF
--- a/.github/workflows/build-types.yml
+++ b/.github/workflows/build-types.yml
@@ -1,0 +1,19 @@
+name: Test Type Generation from Schemas
+
+on:
+  pull_request:
+
+jobs:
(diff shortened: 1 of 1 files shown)


NiklasRentzCAU (approved):
:+1:

PR #159, opened 2023-10-18
Dark theme support

NiklasRentzCAU:
Implements support for dark / other color themes for clients over the LSP and adds a property to be read by syntheses to conform to some theme-related foreground/background/highlighting colors.

Relates to kieler/KLighD#175

PR #210, opened 2024-12-20
Dark theme support

NiklasRentzCAU:
Implements support for dark / other color themes for clients over the LSP and adds a property to be read by syntheses to conform to some theme-related foreground/background/highlighting colors.

Relates to https://github.com/kieler/KLighD/pull/175
Succeeds #159


NiklasRentzCAU:
@a-sr You may test these changes now with your semantics branch and SCCharts, the changes I wanted are now all included in this PR.

A few things that are not yet 100%, but good enough for this feature to roll onto the master, are the following:
- old servers will throw an exception on the console on theme change in VS Code. This is only aesthetic and does not change the behavior and does not cause the server to crash, so this is fine.
- during initialize, VS Code does not send the current theme colors, but hard-coded colors (currently matching the default VS Code light/dark themes), because VS Code does not offer any API to get the colors of the current theme on the extension side.  The real theme colors are only sent on theme change in VS Code, where I can look into the document and find the corresponding CSS styles to get the real colors (feels a little hacky)

PR #218, opened 2025-04-17
replace depthmap with default Sprotty isVisible behavior

NiklasRentzCAU:
and cleaner checks for smart zoom implementation.

changes option for full detail on relative size to take the maximum instead of the minimum dimension of nodes.

Yet missing re-implementation of modifications for absolute position calculations in top-down layout


Eddykasp (approved):
lgtm

PR #225, opened 2025-06-10
Add pre-defined structural tags

Eddykasp:
Introduces pre-defined structural tags. Common graph-structural properties can be filtered without requiring a synthesis author to add this information to the semantic tags.

Tags can be overridden by synthesis authors if desired. If the tag exists in the properties then that value is used rather than executing the defined behaviour.

The following table lists the new tags and their effects and whether a test has been written yet:
| tag | # | $ | has unit test |
|--------|--------|--------|---------|
| children | true if element has children | number of children | <ul><li>- [x] </li></ul> |
| isNode | true if element is node | *not reserved* | <ul><li>- [x] </li></ul> |
| isEdge | true if element is edge | *not reserved* | <ul><li>- [x] </li></ul> |
| isPort | true if element is port | *not reserved* | <ul><li>- [x] </li></ul> |
| isLabel | true if element is label | *not reserved* | <ul><li>- [x] </li></ul> |
| edgeDegree | true if element is node and has incident edges | number of incident edges |<ul><li>- [x] </li></ul> |
| inDegree | true if element is node and has incoming edges | number of incoming edges | <ul><li>- [x] </li></ul> |
| outDegree | true if element is node and has outgoing edges | number of outgoing edges | <ul><li>- [x] </li></ul>|

### DIFF
--- a/packages/klighd-core/src/filtering/parser.ts
+++ b/packages/klighd-core/src/filtering/parser.ts
@@ -95,7 +95,7 @@ export function parse(ruleString: string): SemanticFilterRule {
     const operatorStack: Array<OperatorNode> = []
     const outputStack: Array<Node> = []
 
-    tokens.forEach((token, _index, _array) => {
+    tokens.forEach((token: string, _index, _array) => {
         if (OPERATORS.includes(token)) {
             while (
                 !(operatorStack.length === 0) &&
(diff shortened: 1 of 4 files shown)


NiklasRentzCAU (approved):
Nice, now let's see them in action!

PR #229, opened 2025-06-17
Added opacity to sidebar.

soerendomroes:
Fixes https://github.com/kieler/kieler-vscode/issues/106

### DIFF
--- a/packages/klighd-core/styles/options.css
+++ b/packages/klighd-core/styles/options.css
@@ -50,7 +50,7 @@
     border-radius: var(--kdc-border-radius-default);
     padding: 4px;
 
-    color: var(--kdc-color-sidebar-icon-primary);
+    color: from var(--kdc-color-sidebar-icon-primary) r g b / var(--sidebar-opacity);
     background: var(--background);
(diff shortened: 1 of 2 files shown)


NiklasRentzCAU:
I do not really like the fact that texts also get transparent with this solution. Maybe only set all background colors to colors with alpha value to keep the sidebar texts at their intended color and not as washed out? Then it may be possible to make the background even more transparent, maybe with .8 or .75 alpha

soerendomroes:
That is not possible, since the variable color does not get reevaluated in that solution if we change the color scheme. I already tried that.

NiklasRentzCAU:
What about this one, does it not reevaluate if you use CSS relative colors? `rgb(from var(--kdc-color-sidebar-background) r g b / 0.75)`?

soerendomroes:
Feel free to try it out yourself. The variable is no longer there one I call rgb.
image

soerendomroes:
Omitting rgb solved this. How about this? I would like t discuss this at the meeting to figure out the correct opacity.

PR #230, opened 2025-06-18
Add css property to limit diagram size by sidebar.

soerendomroes:
Currently, I do not know how to reset this variable properly when closing the sidebar.

Moreover, I seem to be missing some padding here, since the diagram is a little cut-off.

image
image

### DIFF
--- a/packages/klighd-core/src/sidebar/sidebar.tsx
+++ b/packages/klighd-core/src/sidebar/sidebar.tsx
@@ -18,7 +18,15 @@
 /** @jsx html */
 import { inject, postConstruct } from 'inversify'
 import { VNode } from 'snabbdom'
-import { AbstractUIExtension, html, IActionDispatcher, InitializeCanvasBoundsAction, Patcher, PatcherProvider, TYPES } from 'sprotty' // eslint-disable-line @typescript-eslint/no-unused-vars
+import {
+    AbstractUIExtension,
+    html, // eslint-disable-line @typescript-eslint/no-unused-vars
+    IActionDispatcher,
+    InitializeCanvasBoundsAction,
+    Patcher,
+    PatcherProvider,
+    TYPES,
+} from 'sprotty'
 import { DISymbol } from '../di.symbols'
 import { PinSidebarOption, RenderOptionsRegistry } from '../options/render-options-registry'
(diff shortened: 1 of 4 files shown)


soerendomroes:
I added the resize button to the diagram title bar.

Moreover, refresh diagram will now resize to fit if resize to fit is enabled.

I think this feels very nice but feel free to try it out ourself before merging.

packages/klighd-core/src/sidebar/sidebar.tsx
```diff
@@ -110,12 +118,28 @@ export class Sidebar extends AbstractUIExtension {
         // Show or hide the panel
         if (currentPanel) {
             this.containerElement.classList.add('sidebar--open')
+            // Set width of sidebar to maximum width of all opened panels.
+            this.maxWidth = Math.max(this.maxWidth, this.containerElement.clientWidth)
```
NiklasRentzCAU, line 122: This line (and its copy in the else case) can be pulled in front of the if statement to avoid duplicate code

packages/klighd-core/src/sidebar/sidebar.tsx
```diff
@@ -110,12 +118,28 @@ export class Sidebar extends AbstractUIExtension {
...
         } else {
             this.containerElement.classList.remove('sidebar--open')
+            // Reset sidebar width when closed
+            this.maxWidth = Math.max(this.maxWidth, this.containerElement.clientWidth)
+            document.documentElement.style.setProperty('--sidebar-width', '0px')
+        }
+        // Find the canvas element and dispatch an action to re-initialize the canvas bounds.
+        // This works since the previous css property will correctly set the width of the canvas.
+        const canvas = (content.elm as HTMLElement).parentElement?.parentElement
+        if (canvas) {
```
NiklasRentzCAU, line 133: To avoid potential errors and a call of the InitializeCanvasBoundsAction with invalid parameters, this should also check if canvas is of the correct type and contains clientLeft, etc.

NiklasRentzCAU (changes requested):
Very nice change. Does not disrupt any previous behavior and just works as expected. I'll merge this with these minor changes

NiklasRentzCAU (approved):

PR #232, opened 2025-06-19
Add fit command to diagram title menu.

soerendomroes:
While creating the other sidebar options, I found it annoying that the refresh does not do the fit to screen.
Therefore, I added this to the title bar such that I can do fit to screen without having the sidebar open.

This is conforming with how KIELER does it, which additionally has the XX commands in its "title menu".
VS Code
image

KIELER

path849

@Eddykasp @NiklasRentzCAU Do you think that all the buttons should be moved here, similar to KIELER?


soerendomroes:
Superseded by #230

PR #234, opened 2025-07-09
Add documentation to reserved tags

Eddykasp:
- descriptions for reserved tags can now be retrieved
- two new functions provide a list of all reserved tags
- add new tags #childNodes and $childNodes


NiklasRentzCAU (approved):
LGTM.
Where do you imagine this description being used in the framework at the moment?

PR #233, opened 2025-07-07
Code review Ctrl+F for Diagrams Bachelor Thesis

lenahar00:

PR #241, opened 2025-10-08
CTRL-F for diagrams

Eddykasp:
- [ ] close search bar using ESC key
- [ ] double check any unresolved comments on old PR #233


packages/klighd-core/src/search-bar/searchbar-actions.ts
```diff
@@ -0,0 +1,680 @@
...
+        properties: {
+            x: xPos,
+            y: yPos,
+            width,
+            height,
+            'klighd.rendering.highlight': highlight,
+        },
+        actions: [],
+        styles: [],
+    }
```
NiklasRentzCAU: In the long term, this should probably be part of a JS KGraph creation library. I would add a TODO notifying that here.

packages/klighd-core/src/search-bar/searchbar-actions.ts
```diff
@@ -0,0 +1,680 @@
...
+    while (queue.length > 0) {
+        const element = queue.shift()!
+
+        // Remove KText highlights
+        if (isKText(element) && element.styles) {
+            element.properties['klighd.rendering.highlight'] = 0
+        }
+
+        // Remove rectangle highlights from children
+        if ('children' in element && Array.isArray(element.children)) {
```
NiklasRentzCAU: Is this a hacky way to loop through graph elements and renderings, as they both have a "children" element? If so, this seems dangerous to do the same type of loop for both types of elements.

packages/klighd-core/src/search-bar/searchbar-actions.ts
```diff
@@ -0,0 +1,680 @@
...
+        const modelId = HandleSearchAction.currentModel?.id
+
+        /* Handle ClearHighlightsActions */
+        if (ClearHighlightsAction.isThisAction(action)) {
+            removeHighlights(HandleSearchAction.currentModel)
+            // make changes visible
+            if (modelId && this.actionDispatcher) {
+                this.actionDispatcher.dispatch(CenterAction.create([modelId]))
+            }
+        }
```
NiklasRentzCAU: We always center the view if we clear the search highlights? I did not see this happen, but wonder why this should happen in the first place.

packages/klighd-core/src/search-bar/searchbar-actions.ts
```diff
@@ -0,0 +1,680 @@
...
+
+        /* Update highlights to show current result orange  */
+        if (UpdateHighlightsAction.isThisAction(action)) {
+            if (action.selectedIndex === undefined || !action.results || !action.panel) return
+            this.updateHighlights(action.selectedIndex, action.previousIndex, action.results, action.panel)
+            this.actionDispatcher.dispatch(CenterAction.create([modelId]))
+        }
+
+        /* searches for all tags on the model */
+        if (RetrieveTagsAction.isThisAction(action)) {
```
NiklasRentzCAU: Because all of these ifs exclude each other this should be an if else ladder

packages/klighd-core/src/search-bar/searchbar-panel.tsx
```diff
@@ -0,0 +1,731 @@
...
+                            },
+                        }}
+                        on={{ input: () => this.handleInputChange() }}
+                    />
+                    <button
+                        className={`search-button ${this.tagInputVisible ? 'active' : ''}`}
+                        title="Toggle tag search"
+                        on={{ click: () => this.toggleTagInput() }}
+                    >
+                        #
```
NiklasRentzCAU: replacing these with nicer icons would better fit into the remaining UI.

NiklasRentzCAU (approved):
A few comments, otherwise looks good :+1:

PR #256, opened 2026-02-24
schema more micro layout messages

Eddykasp:
Adds KStyle, KAction, KInvisibility, KColoring, all types of placement data


NiklasRentzCAU (approved):
👍

PR #257, opened 2026-02-24
Client-side Micro Layout Implementation

CrazyDiamond-75:
I implemented basic support of client-size micro layout estimation and calculation for point-placed, area-placed `KRenderings`.
Main to-dos after this:
- Implementing `KRenderingRef` retrieval step before micro layout estimation and calculation.

Could be done by writing on the `rendering` property at the relevant stage after extracting the `KRenderingLibrary`. I already implemented methods to handle parts of this.

- Support for grid-placed renderings.

This is pretty complicated, as it might require implementing the original `GridPlacer` class or mimicking its behaviour.

- Support for decorator-placed renderings.
- Writing real unit tests to test each implemented main method.

I've added example JSON from the server syntheses without and with micro layout, to compare both graphs after client and server micro layout.

- Correctly estimating and calculating sizes of `KTexts` and `KLabels`.
- Correctly estimating bounds passed to micro layout estimation as minimal size (i.e. when to inherit which size from which container).
- Calculating insets in micro layout estimation step.


Eddykasp (commented):
did some preliminary look through the stuff. I would go ahead and address the minor things and then merge this into our mka/micro-layout branch so that we can continue to work on this from there.

@CrazyDiamond-75 Thanks a lot for your work :)

PR #266, opened 2026-03-18
Nre/assorted fixes

NiklasRentzCAU:
- This adds a "default title rendering" option that automatically picks the first text rendering in each rendering as the title rendering to be scaled by smart zoom.
- Changes the full detail relative threshold to .1
- do not scale up title renderings that cannot be sensibly scaled up by more than 20%


Eddykasp (approved):
Does this scale up the - in the regions of SCCharts?

NiklasRentzCAU:
If the synthesis defines a custom title rendering (which the SCCharts synthesis does), the default does nothing. I do not know right now what happens if a region has no name, SCCharts may handle this as an empty String or scale up the "+"/"-" label in that case

